PR #58, opened 2013-11-26
Added one linear algebra to new UCDavis problem sets.

hkvinge:
This is a test of how one adds problems. If it works, more will be added.


jwj61:
Closing since this has been around a long time.  Feel free to reopen with contributions put in the Contrib directory.

PR #154, opened 2015-02-25
Initial commit with git

DavidProtas:
Most of these are updated versions of problems I submitted to the NPL several years ago.


jwj61:
I am sure this will be fine, but you mention a previous contribution, and I don't see a CSUN directory.  Are they in a different directory?

DavidProtas:
Is this because the earlier work was submitted using subversion? On the other hand, there is a .svn directory inside the differential_equations directory that contains old versions. I have no idea of how they appeared since there is no such subfolder in the folder on my computer that I uploaded with git.

PR #167, opened 2015-06-04
Added Active Calculus chapter 1 preview activities

djhunter:
I'm coding the Preview Activities for Matt Boelkins' _Active Calculus_ book. Chapter 1 is completed, but needs to be tested on students.


jwj61:
When introducing a new directory to Contrib, it should have a README file (so Westmont/README) to say who's git account can make changes to the files.  That should be easy to add.

Ideally, pg files which use external files (like image files) should have a RESOURCES tag (see the current README at the top of the Contrib directory).  I will add them when I get around to it, but it is even better if you beat me to it.

PR #170, opened 2015-06-14
Accept Alfred several Alfred multivariable calc problems, fix minor errors, move graphic files.

gajennings:
Change status of several Pending Alfred multivariable calculus problems to Accept. Reject some others.
Create a new MLT list led by OPL problem HGM5-16-5-31b-Spherical-integrals.pg and add some Alfred problems and one other OPL problem to it, fix minor errors in two OPL problems, move graphic files associated with accepted Alfred problems to Alfred directories in OPL so they'll be there when the problems are moved.


jwj61:
This is taking some time since there is a lot going on (accept/reject, bug fixing, moving files).

Why are you moving graphics files?  Has the original already been moved but the graphics file was left behind?

gajennings:
John,

Regarding the graphics, I moved them because when I looked at the OPL 
after my previous push I saw that in one case a graphics file seemed to 
be missing that an Alfred problem needed, so it appeared that graphics 
files weren't automatically moving along with the associated problem 
when the script moved the problem to OPL.   The Alfred guys don't create 
a folder to hold a problem and its graphics together, they just include 
them both in the same folder with all the other problems so the 
association between problem and graphics may not be obvious.

Would have been better if I had fixed that by creating folders to hold 
problems and graphics together?

Or better if I didn't try to fix things?

George

On 06/17/2015 11:13 AM, John Jones wrote:
>
> This is taking some time since there is a lot going on (accept/reject, 
> bug fixing, moving files).
>
> Why are you moving graphics files? Has the original already been moved 
> but the graphics file was left behind?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/openwebwork/webwork-open-problem-library/pull/170#issuecomment-112900574>.
>

-- 
George Jennings
Math Dept., Calif. State University Dominguez Hills

jwj61:
No, this looks like it is a matter of timing.  Since your first pull request I built the movement of graphics files into the scripts.  So, moving them should no longer be necessary.

gajennings:
OK thanks, I'll leave graphics files alone next time.
--George

On 06/17/2015 01:12 PM, John Jones wrote:
>
> No, this looks like it is a matter of timing. Since your first pull 
> request I built the movement of graphics files into the scripts. So, 
> moving them should no longer be necessary.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/openwebwork/webwork-open-problem-library/pull/170#issuecomment-112933116>.
>

-- 
George Jennings
Math Dept., Calif. State University Dominguez Hills

PR #177, opened 2015-08-17
Improvements to a couple of problems

dlglin:


jwj61:
When a problem uses math objects, we *do* want the ## MO(1) tag in the problem.  So one of the problems should have it added because of your conversion, and another should have it put back in.

dlglin:
Sorry, must have been updating on an out-of-date copy.  Tags are fixed.

PR #182, opened 2015-09-02
Correct small error in display macro.

boshuk:
Change (slightly) the definition of
dispSyst so that a hardcopy will be
generated without an error. (In two
files change ^ to $CARET for the
same reason.)


jwj61:
The change to dispSyst looks fine.

The change of ^ to $CARET seems right outside of math mode, but not right inside math mode.  At a minimum, \( ... 10$CARET {-3} ...\) will render differently in html and hardcopy.

boshuk:
Yes, that was a stupid oversight, now fixed.
Thanks.

PR #183, opened 2015-09-02
Prevent reloading of page when an AnserFormatHelp link is pushed

Alex-Jordan:
This probably needs Paul Pearson's approval.

Currently if you click an answer format help link, along with opening the popup window with the link, the whole page is reloaded.

By inserting `return` as in this commit, once the onClick is triggered and one of the openhelp... functions is run, then since the function ends with `return false;`, it prevenst the `href` from being used.

I can imagine that this is mildly beneficial for general WeBWorK use, in preventing pages from needless reloading. But it's possibly essential for anonymous WeBWorK calls inside iframes, as prototypical examples here demonstrate: http://spot.pcc.edu/~ajordan/ww-mbx/html/section-1.html
Right now, clicking an answer format help link causes an error, but I have tested that enacting this change will make the answer format help links clickable without error.

### DIFF
--- a/OpenProblemLibrary/macros/FortLewis/AnswerFormatHelp.pl
+++ b/OpenProblemLibrary/macros/FortLewis/AnswerFormatHelp.pl
@@ -138,7 +138,7 @@ $angleHelpExists = 1;
 if (!$customstring) { $helpstring = "help (angles)"; } else { $helpstring = $customstring; }
 
 if ($main::displayMode ne "TeX") { 
-    return htmlLink( "#", "$helpstring","onClick='openhelpAngles()';");
+    return htmlLink( "#", "$helpstring","onClick='return openhelpAngles()';");
 } else { return $helpstring; }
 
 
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
@paultpearson should weigh in on this.

paultpearson:
If I understand correctly, replacing "onClick" by "return onClick" stops an error message from being generated, otherwise functions the same, and further enables things to work inside iframes.  If that's the case, then I'm all for the changes.  If that's not the case, then I would to know more details.  Also, John Jones should approve (and then comment and / or commit) these changes since he has been involved in updating this macro.

Thanks!

Paul Pearson

Alex-Jordan:
I _believe_ that the only change in behavior with regular WeBWorK use is that the problem page will not reload upon clicking the help link. I only know all that I know about this stuff from working on it today though, so I don't consider myself to have expert knowledge. But this is after reading up on `return` to research this.

If you go to the link I posted above, it's using a course where I just put a local copy of AnswerFormatHelp where I enacted this for the numbers link help in the first problem, but not for the formulas help link. So you can see the difference in behavior there.

The other thing to test is, if you pull this, does it affect regular use of AnswerFormatHelp. I tested that, and saw no issues, but of course I can only test using my WeBWorK server (which is on develop), my OS (mac OS), and my browser (Firefox). Still, it sounded to me from reading today that if `onClick` is supported by a browser in the first place, then use of this `return` should be fine.

PR #184, opened 2015-09-04
Adding LA Tech Open-source Problems for Engineering Studies to Contrib

LATechOPES:
Open-source Problems for Engineering Studies (OPES) at Louisiana Tech University has created problems for Electrical Circuits, Thermodynamics, and Statics that we would like to be included in the NPL.


jwj61:
The general convention in Contrib is for the top level directory to be the name of the school.  Here, it seems that LATech might be a reasonable choice here.  You can do this with one command

git mv OPES LATech

followed by a commit and and a push.

There is other contribution information in Contrib/README on submitting contributions which you should look at.  For example, the readme file in this commit seems to be empty and it should have some information.  Problems which use extra resources like png images should have RESOURCES tags.  Finally, to the extent it is reasonable, one should use DBsubject/DBchapter tags which are in the current taxonomy.  I realize many of these problems will be in new territory, but there is currently a DBsubject of "Electric circuits" which seems close to "Electrical Engineering and Circuits" which is used here.  Then again, maybe they are very different (I admit I am a layperson for that subject).

jwj61:
I merged this, but in addition to the issues above, there are missing macro files.  They should be put in the repository in webwork-open-problem-library/OpenProblemLibrary/macros/OPES, like

interpMacros.pl and SI_property_tables.pl

PR #196, opened 2015-10-05
Clarify wording on FortLewis Fourier series problems

dlglin:


paultpearson:
Good addition.  Thanks, Danny!

PR #199, opened 2015-10-11
Remove "s" from "derivatives" because first derivative is only answer.

Dick-Lane:


Dick-Lane:
Remark about "First sentence" is about revision of Michigan/Chap3Sec5/Q47.pg

PR #202, opened 2015-10-19
Add UBC problems

xcompass:


jwj61:
It does not seem to be in this pull request, but I just noticed that the UBC stat problems have a directory with a space in the name.  That is not allowed.  I removed the space, and if you do

  git pull upstream master

your paths should be fixed.

xcompass:
@jwj61 Thanks and updated! What's the process to have them included in the official OPL? Is there anything else we need to do?

jwj61:
No, people will look at problems when they can.

PR #227, opened 2016-01-10
Adding STAT344 and ECON325 problems to Contrib/UBC

xcompass:
More problems from UBC


jwj61:
Pull requests should only have changes to problems in Contrib and this also has changes to files under OpenProblemLibrary.  If you do want changes under OpenProblemLibrary, it should be a different pull request and have an explanation as to the changes since those changes are actually reviewed whereas you are free to make changes to your own problems under Contrib.

xcompass:
@jwj61 Thanks for catching that. I've updated the PR to remove the changes in OPL directory.

One question though, it seems only some of our problems are processed to OPL directory from last batch. For example, in Contrib/UBC/SPPH/SPPH400/hw02/, hw02_q01.pg - hw02_q03.pg are processed, but not hw02_q04.pg -hw02_q06.pg. I'm wondering if they are rejected or there is a bug in the processing script that left them out. I attached the full [missing_list.txt](https://github.com/openwebwork/webwork-open-problem-library/files/85372/missing_list.txt)

jwj61:
Sorry for taking so long to get to this.

When I look at your repository, whole directories, like Contrib and Pending are missing.  That means that if I merge your pull request, it will delete all of those files from the main repository.

xcompass:
@jwj61 default display branch on our repo is ubc. This PR is merging from ubc:master, which is the sync branch with upstream (yours). And if you look at the file changed tab, only additions, there is no deletion.

jwj61:
I was looking at the wrong place.  I think I was now loing in the right place.  Files generally seemed to be there, but there were differences with many files which were not in Contrib/UBC.  So, I copied the Contrib/UBC files from ubc:master to the main OPL repository.  In other words, it should have the desired affect, but did it without merging everything in the repository.

PR #231, opened 2016-02-10
added METU-NCC Diff Eq and Linear Alg problems.

benjaminwalter:
eigenvalue / eigenvector problems (including generalized & complex eigenvectors)
laplace transform problems; systems of differential equation problems; fourier transform problems


jwj61:
I do not see a README file for this directory.  Please add one which says who is allowed to make changes.

PR #239, opened 2016-04-01
Update Sage Embeds

drjt:
Using threes to allow 3D graphs to render. Also, adding pretty_print to
remove error messages for older html usage.


drjt:
Updating existing sage 3d graphing problems

jwj61:
It looks like you are changing OPL tags in one of the files.  Unless the old tags were wrong, please fix

drjt:
I'll check.  I thought I grabbed a library version to edit and did not
change any of the tags. Perhaps I got an old version on one of them.

JT

via a Galaxy S6 far far away.

drjt:
Tag fixed on problem.

PR #249, opened 2016-08-23
Saint Michael's contributions to OPL for Linear Algebra

enr0n:
I would like to request that these problems be put into the OPL. There are similar problems for linear algebra, but these are designed to be quick concept-checkers, rather than long homework sets. More importantly, these are designed to support the freely available textbook _Linear Algebra_ by Jim Hefferon. Currently there are no problems to go along with this text.


jwj61:
Please add a readme file to your directory specifying who (what github accounts) are allowed to make changes.

PR #265, opened 2016-11-02
Correct answer

mikeshulman:
Either I'm very confused, or the answer to this problem is wrong.  To get a vector field whose magnitude is inversely proportional to the nth power of r, you have to divide r by the (n+1)st power of its magnitude: r/|r| to get a unit vector, and then (r/|r|)/|r|^n to make the magnitude 1/|r|^n.

### DIFF
--- a/OpenProblemLibrary/Hope/Multi2/12-01-Vector-fields/Vector-fields-07.pg
+++ b/OpenProblemLibrary/Hope/Multi2/12-01-Vector-fields/Vector-fields-07.pg
@@ -38,7 +38,7 @@ $n = random(2,3,1);
 $s = random(-1,1,2);
 @direction = ('toward','away from');
 
-$answer = Compute("<$s*x/(sqrt(x^2+y^2))^$n, $s*y/(sqrt(x^2+y^2))^$n>");
+$answer = Compute("<$s*x/(sqrt(x^2+y^2))^($n+1), $s*y/(sqrt(x^2+y^2))^($n+1)>");
 
 
 ##################################


paultpearson:
Hi Mike,

I wrote this problem a number of years ago.  Upon careful reading, I am convinced it's not a very good problem for several reasons.

1. The problem says (1) "a vector field whose magnitude (i.e., |F|) is inversely proportional to the nth power of |r|" when it should instead say something like (2) "a vector field (i.e., F) that is inversely proportional to the nth power of |r|".   The vector field F is a function of r, so statement (1) would mean |F| = r / |r|^n, which is nonsense because r = <x, y> is a vector while |F| and |r|^n are real-valued.  Instead, we want something like F = r / |r|^n, which comes from statement (2).

2. Generally speaking, if y is inversely proportional to the nth power of x, then y is directly proportional to 1/x^n and thus y = k/x^n for some constant k.  This vector field question gives insufficient information to determine what the value for k should be.  This oversight on my part makes the given answer to this problem incorrect because it is an incomplete answer.

3. I do understand your argument and find it to be a valid interpretation of the problem as originally stated.  (But the original problem statement is flawed, as noted in item 1 above.)

When I wrote this problem, I wanted students to make distinctions between F(r) = r, F(r) = -r, and also realize that they could shrink the vectors in the vector field by F(r) = r/|r|^n or F(r) = -r/|r|^n.  I did not intend for students to have to think about whether it should be |r|^n or |r|^{n+1} in the denominator.  I'm not convinced this problem is very good, and it may just be a frustrating exercise for students (and teachers!).

Do you have any suggestions for improvement?  If not, I'm seriously considering deleting this problem.

Best regards,

Paul Pearson

mikeshulman:
re: item 1, what the problem actually says right now is that the magnitude is inversely proportional to some power of the distance from (x,y) to the origin.  I don't find that problematic, since a magnitude and a distance are both scalars.  Item 2 is true, but it could also be fixed by changing the answer checker, right?

If a problem is deleted from the OPL, what happens to people who have it in a problem set when they update their OPL?

jwj61:
I haven't looked at the problem and leave what is appropriate mathematically to Mike and Paul.  From the perspective of OPL maintenance:

If the problem is deleted, then those problem sets will be broken.  If the problem is badly broken and too hard to fix, then that does not seem so bad.

Obviously the best solution is if the problem can be fixed.  If a complete fix doesn't seem possible, but some of the problem is ok, a fix might mean keeping part of the problem and deleting the unfixable parts.  

If there is nothing to salvage, I suppose a placeholder "problem" could be put in place which essentially produces a message saying that the problem has been removed from the OPL (with the ZZZ tags so that it no longer appears in the library browser).  That might be better than having instructors/students seeing webwork throw generic error messages.

paultpearson:
Mike, you're right.  I just wasn't thinking clearly enough about it at the time I wrote my reply (apparently, I should have eaten dinner before replying!).

PR #266, opened 2016-11-03
Specify a value to determine the proportionality constant

mikeshulman:
As noted by @paultpearson on #265 , this question also doesn't have a unique correct answer until we determine the proportionality constant.  It would be nice to be able to mark any answer correct as long as it's proportional, the way we can check for "any antiderivative" with `upToConstant`, but I couldn't figure out how to do that.


mikeshulman:
Two years later... anyone willing to merge this?  @paultpearson ?

PR #267, opened 2016-11-08
Update DE problem with better answer checking and instructions.

dlglin:
…tion includes both sides of the step function, and add instructions on how to enter the step function.


paultpearson:
Nice use of local contexts.  I wish I had known how to do this a long time ago.

PR #297, opened 2017-03-23
fix several problems in OpenProblemLibrary/Mizzou/Algebra/functions_e…

heiderich:
…valuating/two_graphs_equal_inequal_02.pg

I think the following should still be improved: The problem expects "discrete" sets such as {0,5} to be entered as a list "0,5", while unions of intervals are supposed to be entered as such. I propose that the problem should expect discrete sets to be entered as such and not as lists.


jwj61:
Followed your suggestion on the parts which are discrete sets, and added text to the problem on how to enter the answers.

PR #308, opened 2017-05-17
upload author training files from Rochester

mgage:
Added contrib files used for training in using drag and drop questions.

Some changes made while training on updating tags.


jwj61:
Why do you want webwork-open-problem-library/OpenProblemLibrary/270/setDerivatives1/c1s5p8b.pg to show up in the library browser when it is a duplicate of another problem?

jwj61:
This pull request had two parts, new files to Contrib/Rochester and changes to OPL files.  The latter was causing a conflict, so I added the Contrib files and am closing this pull request.  If there are changes to be made to regular OPL files, it should be based off of current versions.

PR #319, opened 2017-06-26
220 Self Test Matrices

bennylojf:
Please consider these problems for the OPL. These problems are for the mechanical engineering program at UBC, which has a very unique 2nd year. Thus, the problems for this program are unlike anything currently in the OPL.


jwj61:
This pull request changes 55,502 files.  It cannot be merged unless it affects only files under Contrib/UBC

bennylojf:
Hi @jwj61 , I don't know why that happened. I followed the instructions given here: http://webwork.maa.org/wiki/Submitting_problems_to_the_OPL#.WVGdfmgrJPY. When I was using Git, I only added the one folder that I intended to push. Do you know why I changed 55k files when I only added one folder to Contrib/UBC?

jwj61:
Did you delete some of the existing problems?

bennylojf:
@jwj61 , no, all I did was go to the UBC subdirectory, make a new folder called "MECH 2", and copy and pasted the new files into the folder. I then added MECH 2, committed, and pushed it. I did not touch anything else.

jwj61:
I have no way of knowing what went wrong.  I suggest making a fresh clone, adding the directory and files, commit, push, and pull request.

Also, do not use spaces in names of files or directories (folders).

bennylojf:
I will try that, thanks

PR #323, opened 2017-07-17
Added previews 7.4, 7.5, 7.6, 8.1, 8.2

djhunter:


jwj61:
Problems don't get added directly to the OPL problem authors.  They go into Contrib.  If you want them considered for addition to the OPL, make that an issue.

PR #331, opened 2017-08-06
Add Hope problems for calculus 1

paultpearson:
Here are some new calc 1 problems for your consideration.


paultpearson:
I forgot that git doesn't like multiple commits.  I'll assemble all my files (more to come) and submit just one pull request.

jwj61:
I don't know what you mean by that.  If you make a pull request, commit new changes and push it, they are automatically part of the same pull request.

paultpearson:
That's OK.  Just ignore my remarks.  (I still struggle to make git do what I want.)

PR #333, opened 2017-08-07
Specify that x should be a real number (any field of characteristic not equal to 2 would be fine)

heiderich:
The reason is that 4x=8 is not equivalent to x=2 in characteristic 0.


heiderich:
Of course I meant to write: "The reason is that 4x=8 is not equivalent to x=2 in characteristic **2**."

PR #336, opened 2017-08-08
Negate macro

heiderich:
use the negate() function in more of the SDSU problems, in one ASU problem and several Rochester problems
one SDSU problem uses niceTables.pl now


heiderich:
I forgot to mention it in this pull request: It would be good if this was merged at the same time as the pull request 

https://github.com/openwebwork/webwork2/pull/801

to prevent the affected problems not finding the `logicMacros.pl` file.

Could someone please review it?

PR #338, opened 2017-08-15
add subroutine "implies"

heiderich:


jwj61:
Is there a reason to run this through a macro?  In a pg file one can just use \implies, as in \\( P \implies Q\\).

heiderich:
Yes, there is a reason: To get rid of hardcoded things like \implies in problems.

With a macro every instructor can choose the symbol he prefers (other common symbols are $\Rightarrow$ and $\rightarrow$). This makes it much easier for instructors to adopt problems to his notation. Otherwise he would have to edit each problem separately.

I think such kind of macros would make sense in other situations as well. I was inspired by the macros in `OpenProblemLibrary/macros/UMass-Amherst/algebraMacros.pl`.

Are there arguments against it?

jwj61:
I didn't realize people liked those other options.

I am not opposed to the idea, but if more and more symbols are going to be handled this way, it seems that they should be done through a file in pg/macros, maybe in PGbasicmacros.pl so it is always loaded.  There are already macros in there for specific typesetting needs, although typically they are mode specific.  

@mgage what do you think?

heiderich:
The important point for me is to allow the instructor to easily adopt these macros. Having them in `logicMacros.pl`, a copy of this file could be placed in the course directory to overwrite these settings. Would the same work with PGbasicmacros.pl? I would guess that PGbasicmacros.pl will change more from time to time. So overwriting it would imply the need to merge things in recurrent courses. Having small macro files, each for a specific use / area of mathematics, would probably cause less problems. That's my two cents.

paultpearson:
I'm in favor of having small macro files that apply only to a small number of pg files, rather than globally defining things in `PGbasicmacros.pl`.  Also, I'm in favor of having loading the `PGcourse.pl` macro last in each pg file so that instructors can redefine such subroutines at a local (course) level.  Using `PGcourse.pl` also means that authors would not need to determine that the `logicMacros.pl` file contains the subroutines that they want to change.

mgage:
PGbasicmacros.pl is a bit unwieldy -- we've already taken the answer macros out of that file.
Perhaps we could add the logicMacros.pl and to pg/macros/   and then call it from PGstandard.pl?

jwj61:
I see the point about wanting to have a file which could be put as a whole in the search path, making PGbasicmacros a bad choice.  

Paul's suggestion makes sense, but philosophically it seems bad to have to put something in 30,000 files when you know you always want it loaded.  The question is whether webwork's framework provides a reasonable way of accomplishing this.

I like Mike's suggestion, although with the tweak of changing the name of the file to something like customizeLaTeX.pl, or something to allow it to contain things other than logic symbols without confusion.  It also takes care of the problem of needing to explicitly adding the file to loadMacros statements.  I know the contents of some macro files are preloaded by webwork, so presumably adding a new version to the search path of those would not get picked up.  Does this apply to PGstandard?

heiderich:
I do not know well enough the purpose of the standard macro files such as `PGbasicmacros.pl` and `PGstandard.pl` and `PGcourse.pl`. For the sake of easy maintenance I still think it would be more practical to have several small files for certain subjects (such as logic here) instead of one big one such as `customizeLaTeX.pl` in order to keep the part that is overwritten by a locally modified version of these files as small as possible.

I doubt one would need `logicMacros.pl` in 30.000 problem files. But it would probably be useful in many problems. If we load some of these macro files by default, the question will arise which ones to load automatically. Will we load `logicMacros.pl`? Will we also load something like `OpenProblemLibrary/macros/UMass-Amherst/algebraMacros.pl`? I think the latter is very specific and I think it would be okay to leave it to the author to load such macros specifically.

@paultpearson:  Could you please explain in more detail the following? Do you mean to load `logicMacros.pl` in `PGcourse.pl`?

> Also, I'm in favor of having loading the PGcourse.pl macro last in each pg file so that instructors can redefine such subroutines at a local (course) level. Using PGcourse.pl also means that authors would not need to determine that the logicMacros.pl file contains the subroutines that they want to change.

dlglin:
If the goal is to be able to have different behaviour in a problem without editing the problem file, then wouldn’t it make more sense to have an environment variable set that is simply read by whichever macro file needs it?  In this case I’m thinking something like $pg{impliesSymbol} (that might be the wrong variable name.  I’m working from memory) which would be set in defaults.config, and then possibly overridden somewhere down the pipe.

The main advantage of this is that it solves a problem already mentioned, namely that if one makes changes to a local version of a macro file, they will lose out on any updates that are made to the global macro file.  It also prevents having to worry about multiple instances of the same file.  Another thought is that the same configuration could be used by more than one macro file.

This approach also gives you the ability to make the change at either the course level (in course.conf) or at the server level (in localOverrides.conf).

Of course I haven’t really looked at how easy this would be to implement.  There are a couple of other places where I would like to see the same functionality.  The first one that comes to mind is the default display of vectors (i.e. having all Vector objects default to column vector format).


On Aug 15, 2017, at 4:47 PM, heiderich <notifications@github.com<mailto:notifications@github.com>> wrote:


I do not know well enough the purpose of the standard macro files such as PGbasicmacros.pl and PGstandard.pl and PGcourse.pl. For the sake of easy maintenance I still think it would be more practical to have several small files for certain subjects (such as logic here) instead of one big one such as customizeLaTeX.pl in order to keep the part that is overwritten by a locally modified version of these files as small as possible.

I doubt one would need logicMacros.pl in 30.000 problem files. But it would probably be useful in many problems. If we load some of these macro files by default, the question will arise which ones to load automatically. Will we load logicMacros.pl? Will we also load something like OpenProblemLibrary/macros/UMass-Amherst/algebraMacros.pl? I think the latter is very specific and I think it would be okay to leave it to the author to load such macros specifically.

@paultpearson<https://github.com/paultpearson>: Could you please explain in more detail the following? Do you mean to load logicMacros.pl in PGcourse.pl?

Also, I'm in favor of having loading the PGcourse.pl macro last in each pg file so that instructors can redefine such subroutines at a local (course) level. Using PGcourse.pl also means that authors would not need to determine that the logicMacros.pl file contains the subroutines that they want to change.

paultpearson:
Case 1: If a pg file has the line `loadMacros("PGstandard.pl", ... ,"logicMacros.pl","PGcourse.pl");` then if an instructor wants to redefine how the implies symbol is displayed, they can use the File Manager to go to the course `templates/macros/` directory, create a file `PGcourse.pl`, and add some code to the file to define how the implies symbol is displayed.  A subroutine named implies in `logicMacros.pl` will be superseded by a subroutine named implies in `PGcourse.pl` because the `PGcourse.pl` macro is loaded after `logicMacros.pl`.

Case 2: Another alternative would be to use `loadMacros("PGstandard.pl", ... , "logicMacros.pl");` and then have the instructor copy the file `logicMacros.pl` to their course `templates/macros/` directory and edit it to change how the implies symbol gets displayed.

In both cases, we're using the fact that any macro files in the course `templates/macros/` directory will supersede another copy of the macro file in a different location.

When I said, "Using PGcourse.pl also means that authors would not need to determine that the logicMacros.pl file contains the subroutines that they want to change," I meant that if an instructor wanted to change how the subroutine named implies was defined but did not want to look through each of the macros listed in the `loadMacros()` command to find which one or ones had code for the subroutine named implies, they could instead just define the implies subroutine in `PGcourse.pl` and everything would work.

There may be better methods for solving this problem than the ones I am suggesting, and I would like to know them.

heiderich:
One additional consideration: Certain constructions require arguments (for instance the groups in `OpenProblemLibrary/macros/UMass-Amherst/algebraMacros.pl`). In principal the subroutines defined here should also take arguments. The function `negate` is 1-ary and `implies` is 2-ary. We may consider to pass arguments to these functions. For instance we may want to use (modulo possible quotations marks etc.)

`negate(p)` instead of `negate() p`

and

`implies(p,q)` instead of `p implies() q`.

Why do you think about this?

jwj61:
On the suggestion of arguments, I would not require arguments unless it is really needed. I think it makes it unduly cumbersome.

I tried implementing a combination of Paul's and Mike's suggestions.  I

- created a new macro file for latex macros, and put the logic, set theory, and algebra macros in it
- made PGstandard.pl load that file
- made a problem which only loads PGstandard.pl, and then PGcourse.pl which uses implies()

Then everything worked as desired, including when I redefined the macro I wanted in mycourse/templates/macros/PGcourse.pl.

Any comments before I make a pull request to webwork/pg?

It seems the hard part will be making all of the OPL problems load PGstandard first and PGcourse last, but that shouldn't be too hard.

mgage:
sounds ok to me.

Mike

jwj61:
I just changed everything in the OPL to load PGstandard.pl first, and PGcourse.pl last.  They also don't load the files which PGstandard loads.  The first downside is that comments on lines after a macro file were not kept, but I think that is minor.  Commented out macro files do not appear at all (even more minor).

While I tried to think of everything, and the diffs I looked at seemed right, I probably didn't, so prepare for disaster to strike and to report it so the changes can be fixed or rolled back.

jwj61:
This is now in customizeLaTeX

PR #340, opened 2017-08-18
add setTheoryMacros.pl, a file that should collect macros related to set theory

heiderich:
The purpose of this macro is similar to that of `logicMacros.pl`.


jwj61:
These macros are now in customizeLaTeX

PR #381, opened 2017-10-20
Corrected for negative radius answers

adentremont:


adentremont:
@xcompass Already closed - no rush on this one

PR #385, opened 2017-10-30
Update hw06-q2.pg: remove those non-ascii character

chenhao6459new:
remove those non-ascii character requested by @xcompass


xcompass:
Thanks 👍

PR #407, opened 2018-01-15
remove the variable "x" from the context

heiderich:
By removing the variable "x" from the context an error message is shown
to the student in case "x" is contained in the answer string (for
instance if he writes "x" instead of "t").


OpenProblemLibrary/NAU/setLinearAlgebra/minpoly2.pg
```diff
@@ -26,6 +26,7 @@ TEXT(beginproblem());
 
 Context('Numeric');
 Context()->variables->add(t => ['Real']);
+Context()->variables->remove("x");
```
dpvc: There is a one-step means of doing the two lines above:

```
Context()->variables->are(t => ['Real']);
```

(changing `add` to `are`).

heiderich:
Thanks. I updated the branch.

PR #410, opened 2018-02-04
Mech Eng taxonomies

adentremont:
Adding suggested taxonomies (V1) for mech eng subjects.  (Will send confirmed version to John Jones to consider for inclusion in OPL taxonomy in the coming week. More details then).


jwj61:
Our process in the past has been to get faculty together from different universities together to decide on the taxonomy.  We will have a small workshop for people to do just this sort of thing in late June.  If there are people from your university who might be willing to participate, can you e-mail me (jj@asu.edu) their names/e-mail addresses?

PR #412, opened 2018-02-22
New folder for BCIT

jlo99:
We would like to start contributing to the OPL.


jwj61:
The readme should also say which github logins are allowed to make changes to this folder.

PR #433, opened 2018-06-02
Add gianni to UBC readme

adentremont:
I'm not sure this is right, but I can't seem to create a pull request for adding Gianni to our readme without the commits below where I updated our fork.  If it's not right, I will try to fix it (suggestions welcome).


jwj61:
This pull request has a conflict.  It is such a small change, I would suggest closing it, and re-pulling from github, and then doing it again.

PR #441, opened 2018-06-29
Adding and updating problems

gianni-co:
Includes new and old problems with updated folder structure.

- Folder structure completely redone and every problem has been renamed with a unique ID 
- Old Dynamics and Solid Mechanics questions have been updated with the subject-chapter-section in the header
- Every file that has been identified as part of a Mech2 pre-lab or problem set also had their respective keyword lists updated.


jwj61:
In the future, do not put spaces in directory (folder) or file names.  I fixed the ones here, so be sure to pull from upstream master before doing anything else.

gianni-co:
Thanks for approving my pull request. I'll keep that in mind for my future
commits.

PR #483, opened 2018-11-09
Fix bug in optimization problem.

d-torrance:
The y-coordinates of the solutions involving squaring a number, but the
problem was using "^2" instead of the Perl-native "**2".  Since it was using
num_cmp to check the answers, it was marking the correct answers as wrong.


jwj61:
This fix was correct, but it motivated me to convert this problem to use mathobjects, and included your fix in that.

PR #496, opened 2018-12-29
CC Denver problems

Alex-Jordan:
I am facilitating Community College of Denver donating their problems to the OPL. These problems are mostly from the OpenStax College Algebra textbook.

I've asked James Morski, the lead from CCD, to create an account on GitHub and I will ping him here.

I'll be following up with some commits to tag problems, including where they come from in the OSCA book. Is it wise to make those commits here in Contrib? Or later after these problems are in the OPL?


jwj61:
Doing tagging here is good.

PR #522, opened 2019-02-04
Make it clear many row reductions are possible

kcrisman:
I had two different complaints about this problem after I made it clear in class "you can row reduce any legitimate way you like" and then this problem makes it sound like there is only one way to do it.  (Naturally, there might be more efficient ways than others!  But I think my students' point is well taken.)  I hope this is a minimal change that is acceptable to propose.


kcrisman:
Also, separately there seems to be a typo "and are" where the word "and" is extraneous.

PR #531, opened 2019-02-21
contributing problems for reading Vernier scales

duffee:
I'd like the problems considered for the OPL.

There aren't as many physics questions in the OPL and no lab skills or experimental questions that I've seen.  I've suggested what I consider appropriate subject/chapter/section headings.  If there's someone out there with a Physics subject ontology that you like, let me know.

Happy to coordinate work with anyone wanting to improve the coverage of physics (and maybe chemistry) problems.


duffee:
next commit adds some questions on nuclear particles

PR #555, opened 2019-04-19
Adding 2 new circuits problems.

gianni-co:


jwj61:
Please don't add problems directly to Pending.  When we have people ready to review problems, they will be copied there automatically.

jwj61:
This still seems to make changes to Pending

PR #563, opened 2019-04-30
Correct answer was transposed

jirilebl:
So if $i == 0, this might have gone unnoticed, since the vectors were [[1],[1]] and [[1],[4]] so transposing the matrix [[1,1],[1,4]] did not change things unless the student wanted to enter [[1],[1]] and [[0.25],[1]] which then wouldn't work.  If $i==1, then this never worked, in fact, in that case, if you enter the vectors [[1],[4]] and [[1],[1]] as given by the "correct answer" it marked you incorrect.

One thing that bothers me is that it seems that by the usage statistics in the library browser that many students (more than half) got through this.  Did students just start randomly permuting the numbers they got until the problem let them through?

Anyway, for that reason it may be good for someone to double check me.  Try especially entering some other multiple of the eigenvectors so that a transposition gets caught.

### DIFF
--- a/OpenProblemLibrary/FortLewis/DiffEq/3-Linear-systems/04-Eigenvalue-method/KJ-4-4-13-multians.pg
+++ b/OpenProblemLibrary/FortLewis/DiffEq/3-Linear-systems/04-Eigenvalue-method/KJ-4-4-13-multians.pg
@@ -91,7 +91,7 @@ $v22 = Real("1");
 }
 
 
-$multians = MultiAnswer($v11, $v12, $v21, $v22)->with(
+$multians = MultiAnswer($v11, $v21, $v12, $v22)->with(
 
     singleResult => 1,
     checkTypes => 0,


paultpearson:
Hi Jiri,

Thanks for taking the time to go through these problems carefully.  I wrote these webwork questions during the semester I taught DE's, so it's not surprising that they contain a few mistakes.  I don't know how we can get better reporting of bugs without educating more instructors on how to use either github issues or the webwork bugzilla reporting system.  It is unfortunate that so many people just live with a bug rather than reporting it.  If you ever suspect that a bug has propagated to many questions (e.g., one problem with a bug was used as a template for other problems, thereby spreading the bug) please let me know.

Thanks!
Paul Pearson

paultpearson:
If in the answer checker we print the student inputs by adding the line warn @s;

	my @s = @{$student};
        warn @s;

and then input an answer (say 1, 2, 3, 4) and click Submit Answers, we can see that the student answers appear in the order upper left, lower left, upper right, lower right (i.e., column major format).  That's because the answer blanks are being generated in that order when the html source is created.  So, the multianswer() function needs to harvest the student answers in the order upper left, lower left, upper right, lower right.  Your bug fix correctly identifies and fixes this issue.

Thanks!
Paul

jirilebl:
Well it might have been that people just did not notice.  I believe I used this in a class (a very small enrollment, perhaps 10 people), and not one student complained, so I haven't noticed it.  Anyway, I think it just seemed weird that so many students manage to get to a wrong answer after a short number of tries.

In my diffyqs set, https://github.com/jirilebl/diffyqs-webwork/, where I'm quite heavily editing the problems, I just replaced the problem with one that uses two matrix entries so that MultiAnswer isn't used, but mainly I made it so that there are more possible answers.  That's actually how I found the issue.

PR #576, opened 2019-08-29
Active Calculus def files

b-stone:
We added `def` files for the webwork problems of the 2018 updated HTML version of Active Calculus.


jwj61:
You aren't listed in the README as someone who can make changes to this directory of Contrib.  The person listed there should have a pull request which adds you first if they want you to modify the contents of Westmont

b-stone:
On line 12 of the README file I am listed as a contributor. Do we need to be more specific, i.e. give a list or something?

b-stone:
Thanks. Should we redo our README file to make it less confusing? I can start it would a list of approved contributors if that would help.

jwj61:
Yes, that would be helpful.

PR #595, opened 2019-09-18
Possible incorrect conditional probability

kcrisman:
I always find these probability problems to be quite tricky.  Hopefully this analysis is correct.

I believe that since this is worded as the conditional probability P(rej|pos), then if this is supposed to be a Bayes' theorem problem we would have to turn it into something involving P(pos) in a law of total probability sense.  That is, P(rej|pos) = P(pos|rej)*P(rej) /  P(pos), where P(pos) = P(pos|rej)*P(rej) + P(pos|not-rej)*P(not-rej).  In words, the probability of a positive test is the sum of the situation where it's positive and actually being rejected and the one where it's positive and not being actually rejected.

The probability of P(pos|not-rej) is indeed `$d`, and `$e` and `$g` are defined to be the complementary probabilities of rejection and non-rejection.  But P(pos|rej) should be `1-$d`, since both of them are about positive tests.  The negative test number hardcoded 20 shouldn't be relevant.

My apologies for putting this all in a commit message, but I didn't want to clutter up GH by putting in a separate issue and PR when I had found the solution, or so I believe.

### DIFF
--- a/OpenProblemLibrary/UVA-Stat/setStat212-Homework04/stat212-HW04-18.pg
+++ b/OpenProblemLibrary/UVA-Stat/setStat212-Homework04/stat212-HW04-18.pg
@@ -41,6 +41,7 @@ $a = random(5,10,1);
 $b = random(25,35,1);
 $d = .01*$a;
 $e = .01*$b;
+$f = 1-$d;
 $g = 1-$e;
 
 BEGIN_TEXT
@@ -70,6 +71,6 @@ $BR
 
 END_TEXT
 
-ANS(num_cmp(($e*0.80)/(($e*0.80)+($g*$d))) );
+ANS(num_cmp(($e*$f)/(($e*$f)+($g*$d))) );
 
 ENDDOCUMENT();       # This should be the last executable line in the problem.


kcrisman:
After some discussion with a probability colleague, I am fairly certain 😄 that this is a necessary change.  Said colleague preferred that the wording of the problem be changed instead to track more with the usual wording of such situations, where one would talk about the percentage of positive results being correct and the percentage of negative results being incorrect, or maybe of both being correct, but that seemed harder for me to envision as very much a non-probabilist.

jwj61:
Sorry for being slow on this, both in the sense of taking a while to respond and in taking longer to understand.

The problem says that $a % (=$d) of the positive tests are incorrect.  I would see that as P(not-reject | pos) = $d, but you have $d as P(pos | not-rej), which of course is different.  Can you explain that point?

kcrisman:
At this point I don't even know what I myself think!    But in that case I don't think `$d` would even be relevant to computing this probability, or?  I think we are trying for computing P(rej|pos).

The way we (me and class) rationalized this was that the thing you know prior is whether or not the organ was rejected, and then you (as a researcher) would go back and see whether the test was positive or not.  Since the other conditional probability you mention should simply be 1-P(rej|pos) = P(not-rej|pos) then the problem would have quite a different answer.  Given (!) that the answer seems to imply using Bayes' Theorem, I guess my interpretation is what was desired, though I agree that the wording is vague, to say the least.

My colleague actually recommended completely removing the problem, and he is actually a probabilist in operations research by training.  But that seemed like a pretty drastic step to me.

jwj61:
I contacted the problem author.  In class, they had explained how to interpret X percent of positive tests are wrong, so indeed, $d = P(pos | not-reject).  But the same interpretation would say P(neg | reject) = 0.2, which means P(pos | reject) = 0.8, which is what is used in the problem.  The value 1-$d is P(neg | not-reject), which does not enter Bayes formula.  So, I am closing this.

My only concern is the $d=P(pos | not-reject) (which still seems crazy to me).  If that interpretation is not standard, then the problem should be changed.

kcrisman:
Okay, I think I see how they are interpreting it.  I think I interpreted half of it your way and half of it their way.  Yikes!  Maybe the problem should be rewritten to be more globally useful, then, but I would leave that to the author.

PR #600, opened 2019-09-20
Answer checkers were not working in part 2

lahvak:
Move the answer checkers, create variables for them and reference them
in the PGML answer fields seems to fix it.

For some reason, the answer checkers in part 2 were all wonky.  The first one kept insisting that three points were not entered, although they clearly were.  The second answer blank has the actual code of the answer checker as correct answer, and so on. 

Moving the answer checkers into separate variables, and using the variables in the PGML answer fields instead the answer checker code seems to fix the problem.  Strangely, part 3 has actual answer checker code in the PGML answer fields, and it seems to work fine.


paultpearson:
Great work!  Thanks for catching this. I had made those same changes locally (within the past few years?), but must've forgotten to make them globally.  Thank you! -Paul

PR #677, opened 2020-03-22
Changed unit vector syntax and added ' to x'y'z'

tleonzio:


jwj61:
Is it ok to be removing the comments at the beginning on copyright, etc.?

tleonzio:
> Is it ok to be removing the comments at the beginning on copyright, etc.?
I'm not sure, that's a good question. I only knowingly made changes to line 61 and onward, I'm not sure why the rest of that text is gone. I can ask my supervisor.

PR #679, opened 2020-03-22
Added clarifying units and changed Ans formula

tleonzio:
Changed answer formula to use different numbers (35 MPa) and added MPa in sigma greater than/less than statement for clarity.


jwj61:
This one also has the big comment removed from the beginning

tleonzio:
Thanks for letting me know. If I want to retain those and just make the changes I originally intended on making, how would I roll back the change so I can make sure the big comment stays in?

jwj61:
I would just go forward.  Make a copy of the old version of the file, switch to this branch, paste the material back in, commit, and push.

PR #724, opened 2020-04-02
Problem Requires Paraphrasing Rewrite

tleonzio:
Problem Requires Paraphrasing Rewrite, Copy has been made to make the changes and the file will be added once finalized. I have about 15 more of these but can't figure out how to combine the commits into one pull request.


jwj61:
In one branch, use

git rm filename

for each file you want to delete (and delete the now unused branches with git br -D badbranch).

Then commit, push, and make pull request.

tleonzio:
Thats on the desktop github client correct? I've just been using the web based github on chrome

jwj61:
The commands would be for the command line interface to git.  Whatever interface you are using, I am sure it is not hard to change more than one file in a given branch.

PR #737, opened 2020-04-22
Removed Problems which require rephrasing

tleonzio:
The problems removed were those which were just removed from Contrib. The mechanics of the problem are not being changed, just the phrasing


jwj61:
You can delete problems from your area of contrib, but not from the OPL.  It should wait until you have an improved version, and then make a pull request to improve it.

tleonzio:
Thanks for the heads up, I'll try and change them ASAP then

PR #746, opened 2020-06-06
adding xinglong-li to contributor

xinglong-li:


jwj61:
You can't add yourself.  Someone who already has permission to make changes in the directory has to make this change.

PR #767, opened 2020-09-05
Use method answerChoice instead of hacking internals

mikeshulman:
The recently merged PR #763 involves some hacking with the internals of the RadioButton implementation, which might break on future updates.  The PR openwebwork/pg#502 adds methods to RadioButton that eliminate the need for such hacking; this PR updates #763 to use these methods.

However, I don't know what the practice is for mantaining compatibility between the OPL and particular versions of webwork.  It seems clear that this PR should not be merged until openwebwork/pg#502 is not only merged but incorporated in a released version of webwork, but perhaps one should wait even longer to give people time to update their webwork installations?  Is there a policy regarding this?


mikeshulman:
openwebwork/pg#502 is now merged.

gajennings:
There's now a simpler way to accomplish this by using the "values" option that's described in the POD documentation for for parserRadioButtons.pl https://github.com/openwebwork/pg/blob/main/macros/parsers/parserRadioButtons.pl lines 53-57 and 113-119.  I followed your idea but used the simpler method to update the code for this problem, so I'm closing this pull request.

PR #783, opened 2020-10-27
Update APEX_3.3_20.pg

pschan-gh:
Not sure if this is a bug or a matter of different conventions adopted by different institutions.  If a function is increasing on (a, b) and on (c, d), can we say in general that it is increasing on (a, b) U (c, d)? I would say no, but others might disagree?


dlglin:
I agree that in general it is incorrect to say that the function is increasing on (a,b) U (c,d).  I believe that the old interval checkers did not support lists of intervals, which is why many old questions accept unions of intervals for when a function is increasing/decreasing.
When updating such problems I think it is important (as you did) to include instructions on how to type a list of intervals.

pschan-gh:
I see, thank you!

PR #793, opened 2020-12-28
Optimization pg files

mathriddle:
Seven standard optimization problems. Each uses scaffolding and includes GeoGebra applet.

### DIFF
--- a/2019JMM/mathriddle/README.md
+++ b/2019JMM/mathriddle/README.md
@@ -1,0 +1,4 @@
+This is a README file for practice
+
+Who does this file belong to?
+Look in Wiki for instructions for README file from John Jones
--- a/2019JMM/mathriddle/fixedAnswer.pg
+++ b/2019JMM/mathriddle/fixedAnswer.pg
@@ -1,0 +1,231 @@
+########################################################################
+
+DOCUMENT();      
+
+loadMacros(
+   "PGstandard.pl",     # Standard macros for PG language
+   "MathObjects.pl",
+   "PGgraphmacros.pl",
+   "AppletObjects.pl",
+   "parserImplicitEquation.pl",
+   "source.pl",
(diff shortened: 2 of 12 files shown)


mathriddle (commented):
Ignore this file

jwj61:
You should remove the 2019JMM files from the pull request.

If I merge this as is, I will delete them, and then when you update with git they will be removed from your copy as well.  If you remove them from the pull request, then you can keep the files in your copy of the OPL.

mathriddle:
John,

Thanks for your message. I'm sorry I messed this up. I didn't realize those
would be included in the pull request until after I submitted it, and I
couldn't figure out how to remove them. They were practice problems left
over from the 2019 JMM workshop. How can I remove them from the pull
request? I have only been using the web access to github. I'm still
learning how git works.

I do have copies of the two JMM files, so if you have to delete them after
doing the merge, that would not be a problem for me.

Larry

jwj61:
Do you have access to the computer having the files in a terminal window?  For example, did you do a git commit at some point?

If so, you should be able to (I have not actually tried this):

git restore --staged 2019JMM/mathriddle/README.md
git restore --staged 2019JMM/mathriddle/fixedAnswer.pg
git restore --staged 2019JMM/mathriddle/rational6.pg

from inside webwork-open-problem-library.  Then try "git commit" (maybe not needed), and then "git push".

If that succeeds, the pull request will update itself.

mathriddle:
Ok, I entered all those commands from a terminal window. Not sure how to
check if the pull request updated itself. Could you let me know if this
worked. If not, just go ahead with the merge where you will delete the
JMM2019 folder. I was planning to delete the folder anyway. I wasn't sure,
however, if I could do this before the merge without messing up the pull
request. Thanks for your help.

PR #795, opened 2020-12-30
GTIIT fixes 2020

taniwallach:
1. A bug fix to `OpenProblemLibrary/OKState/PDE/firstorder/characteristics4.pg`.
2. Improvements to `OpenProblemLibrary/Hope/Multi1/04-04-Basis-and-functions/Matrix_rep_05.pg`:
    - Work around nomenclature differences.
    - Modified parameter generation code to avoid the two C.O.B. matrices from being equal, and to limit number of loops (avoid potential long/infinite loop).
    - This change **will** change some students' version of the problem.


taniwallach:
@gajennings  - Would you mind reviewing the changes?

PR #822, opened 2021-04-22
more circle problems, typos corrected

brittnilorton:


drdrew42 (approved):
Thanks for updating the tags too!

PR #829, opened 2021-05-13
additional problems and bug fixes

brittnilorton:
Additional Algebraic Literacy Problems, and bug fixes for roots with MathQuill

### DIFF
--- a/Contrib/CCCS/AlgebraicLiteracy/Fractions/Add_Fractions_3.pg
+++ b/Contrib/CCCS/AlgebraicLiteracy/Fractions/Add_Fractions_3.pg
@@ -1,0 +1,82 @@
+## DESCRIPTION
+## 
+## ENDDESCRIPTION
+
+
+## DBsubject(Algebra)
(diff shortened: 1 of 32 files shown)


drdrew42 (changes requested):
Looks good, though I found a couple places that need attention.

Also, regarding `mathQuillOpts`, we are changing the way that these options are implemented in the new WeBWorK version (2.16). The documentation isn't fully completed yet, but you should search-and-replace the following:

* search: `->cmp(mathQuillOpts => 'rootsAreExponents: false')`
* replace: `->cmp(mathQuillOpts => { rootsAreExponents => 0 })`

The updated code will not 'work' for problems in WeBWorK 2.15, but after updating to 2.16, the MathQuill options will take effect. The release team is currently working on installation and upgrade instructions, and the 2.16 release should be coming shortly thereafter.

drdrew42 (approved):

drdrew42:
Note to future-self, convert `mathQuillOpts` once 2.16 is released.

PR #831, opened 2021-05-27
Start a SVSU collection in Contrib

lahvak:
This first commit adds all our current College Algebra problems.

They are hopefully correctly tagged.  We would be delighted if any of these eventually ended up in the OPL.


lahvak:
When reviewing the problems, I realize now that while most of the problems were written at SVSU, there are few problems that are modifications of existing problems from the OPL.  The edits are mostly in terminology and notation, which is why we chose to fork the problems and create a local version rather than modifying the existing problems.  I do not know what is the proper way of handling this, but I will at least add some explanations to the modified files.

drdrew42:
> there are few problems that are modifications of existing problems from the OPL

Would you be able to identify which problems are modifications, and perhaps also which problems they came from?

lahvak:
For now I removed all problems that were modifications of existing problems from the OPL.  I will go through those one by one and see if some of them can be submitted as changes to the existing problems instead.

All the remaining problems after the last commit were written here at SVSU.

drdrew42:
Thanks @lahvak, for separating out those problems.

The way that modifications to existing problems have been handled to this point is through the `more-like-this` grouping feature. When you've identified the 'original' problems for each of your modifications, there are a few things that could be done. 

If `Library/original/problem.pg` is the original problem and `Contrib/modify/problem.pg` is the revised problem:
1. Check `Library/original/problem.pg` for an `## MLT(<<name>>)` tag -- _if_ one exists.
2. Copy-paste the MLT tag line from `Library/original/problem.pg` into `Contrib/modify/problem.pg` and skip the remaining steps.
3. If no MLT tag exists, add an `## MLT(<<unique_name>>)` tag line to both `Library/original/problem.pg` and `Contrib/modify/problem.pg` (`unique_name` because we want to ensure that a new MLT group is formed)
4. Add the line `## MLTleader(1)` to `Library/original/problem.pg` *only*.
5. If no MLT tag exists, (and steps 3 & 4 are too much), just add the file path: `Library/original/problem.pg` as a comment inside of `Contrib/modify/problem.pg` (and maybe a second comment to briefly describe what changes were made, e.g. 'terminology: this to that' or 'notation: this to that')

Ultimately, a 'modified' problem should have either
* an MLT tag that is identical to the 'original' problem, or
* a comment that contains the filepath of the 'original' problem with some mention that the problem is a modification

lahvak:
Thank you, this is really useful. Is there any policy on generating the unique MLT tags?

drdrew42:
> Thank you, this is really useful. Is there any policy on generating the unique MLT tags?

Unfortunately, no, not really. I'm working on some tools to help address this though.

However, in the meantime, if you have command-line access, the following command can help confirm that you *do* have a unique MLT name:

`grep -r 'MLT(possibly_unique_name)' /opt/webwork/libraries/webwork-open-problem-library`

For example, if you perform the above action for `MLT(WorkProblem)`, you should see that there are already problems in the OPL using this MLT name, so using 'WorkProblem' would _not_ create a new MLT group.

MLT group names should not include any spaces, use underscore or hyphen or Camel case to deal with multiple words.

drdrew42:
For future reference, I have added these instructions to the [Submitting problems to the OPL](https://webwork.maa.org/wiki/Submitting_problems_to_the_OPL#Submitting_.27similar.27_or_.27modifications.27_to_existing_OPL_problems) page of the WeBWorK wiki.

lahvak:
Sorry, I keep asking questions.  I am now looking at another problem from our collection, and it is a problem from the OPL rewritten using PGML, with added hints and a solution, but with no modification to the actual wording of the problem. Would it be better to submit it as a new problem with the MTL tags, or to create a separate pull request changing the old problem in the OPL?

drdrew42:
> Sorry, I keep asking questions. 

It's really no problem.

If an update doesn't change the main text or logic of the problem, it should be pushed as a replacement for the existing OPL problem rather than as an MLT modification. I view the addition of hints/solutions in the same way as bug fixes -- there's no benefit to keeping the 'old' version. (GitHub keeps track of changes, so the original is never actually 'lost'...)

pstaabp:
Merging this and then adding a PR with minor fixes for these problems.

PR #857, opened 2021-07-22
add ChamplainSaintLambert to Contrib

mtitcombe:
add ChamplainSaintLambert to Contrib in the OPL, and subdirectory Titcombe, and TrigDegrees files


mtitcombe:
By accident, a file in CUNY was added with uppercase S (monoDenom-Scaff.pg) by mistake. The original CUNY problem had a lower case s.

mgage:
Michele -- we guessed wrong.  The file with the capital S is the one that should be preserved -- I'll merge this PR and then I'll fix it with a new pull request.

mgage:
will replace monoDenom-scaff.pg with monoDenom-Scaff.pg in the next pull request

PR #860, opened 2021-08-08
Added 4 textbooks

capacapac:
Engineering Mechanics by Hibbeler (14th ed)
Vector Mechanics by Beer and Johnston (10th ed)
Engineering Mechanics - Statics by Meriam and Kraige (9th ed)
Engineering Mechanics - Dynamics by Meriam and Kraige (9th ed)

I copied the chapter titles from publisher websites and tried to remove special characters (like theta / pi) manually. Hopefully there are none remaining but in case things break, lingering symbols might be the culprit


capacapac:
Updated with new request

PR #864, opened 2021-08-11
updated tagging in AlgLit, added solutions and problems to QuantLit

brittnilorton:
additional updates to AlgebraicLiteracy and QuantitativeLiteracy folders


drdrew42 (approved):
Thanks for updating these tags!

PR #865, opened 2021-08-13
Use `round` instead of `sprintf` to round percentage

d-torrance:
There are sometimes rounding errors due to  floating point issues when the percentage ends in 5,
e.g.:

```
$ perl -e 'printf("%.0f\n",  0.5)'
0
$ perl -e 'printf("%.0f\n",  1.5)'
2
```


drdrew42 (approved):
and your $ans is actually a number now, rather than a string...

PR #870, opened 2021-08-20
Added textbook tags 

capacapac:
I skipped tags on about 30 problems because there's a version mismatch between the existing files on OPL and the UBC repo. I will upload those when that gets resolved.


drdrew42:
can you clarify what you mean by 'version mismatch'?

capacapac:
Sorry for the confusion. There were updates to some of the problems on the UBC repo (reworded problems, removed solutions, etc.) since they were pushed to OPL. 
Since I wasn't aware of the changes, I just didn't modify the updated files here at all. So the only updates I made were adding tags to files that were identical between the UBC repo and this.

capacapac:
Update: Solutions weren't removed, but were added to the OPL version and not synced with the UBC repo, so I'll check each on a case-by-case basis and update the remaining ~31 problems soon.

drdrew42:
Excellent, thank you!

PR #871, opened 2021-08-20
Added textbook tags

capacapac:
And:
updated wording on 4 problems
fixed units on 1 problem


drdrew42 (approved):
Excellent, thank you!

PR #874, opened 2021-08-22
adding example badges to readme

eltenedor:
commit activity shows `invalid`.


drdrew42 (approved):

drdrew42:
shields.io warns that GitHub rate limits some of the data used for rendering these badges... this can be fixed by authorizing the shields.io application directly. I'll see if that helps.

PR #875, opened 2021-08-22
delete test file

eltenedor:
It has been here for 2 years and it does not contain any information. I think one can delete it.


drdrew42 (approved):
😂

PR #877, opened 2021-08-24
Fix answer for poker probability problem.

d-torrance:
When computing the probability of getting at least three of a
particular rank, the numerator will be:

* the number of ways of getting 3 of the given rank and 2 of
  the remaining 48 cards = choose(4, 3) * choose(48, 2)
* plus the number of ways of getting 4 of the given rank and
  1 of the remaining 48 cards = choose(4, 4) * choose(48, 1)

There was a typo in this first term.  In particular:
choose(48, 2) = 48 * 47 / 2 = 24 * 47, not 27 * 47


drdrew42 (approved):

drdrew42:
Thanks!

PR #881, opened 2021-09-09
Fix non-ascii characters

xcompass:


xcompass:
Any idea why only some of the problems are included in OPL?
e.g. hw02_q0[1-3].pg are there: https://github.com/openwebwork/webwork-open-problem-library/tree/master/OpenProblemLibrary/UBC/SPPH/SPPH400/hw02

But not hw02_q0[4-6].pg. Thanks.

drdrew42 (approved):
No, I'm not sure why some of the problems were 'approved' and not others. We are currently brainstorming improvements to the way that OPL problems are reviewed and approved. I would welcome any feedback or ideas on this process.

PR #896, opened 2021-11-03
webwork2.15 requires this to be "Point" context

benwalt:
Typo in description.  This problem worked fine in 2.15, but needed to be changed for 2.16.  
Possibly there will be more corrections like this as the semester progresses....


drdrew42 (approved):
I am still looking into why the default behavior for 'Numeric' context might have changed. It looks as though, historically, the Numeric context has been using open parens as the start of a List object, and not a Point object.

PR #898, opened 2021-11-03
apparently webwork 2.16 is rounding Real to two decimal places?

benwalt:
As far as I know this problem was working fine in webwork 2.15
There is a problem when $ans[2] is a fraction, since it is getting rounded off when being plugged into $ans[1] (in the computation of $ans[3]) yielding too much rounding error for correct answers to be accepted.  Maybe if I didn't cast it to Real right away then it wouldn't have been an issue?  Anyway, I fixed it by manually plugging in the formula for $ans[3].

### DIFF
--- a/Contrib/METU-NCC/MAT100/SE1/Line.pg
+++ b/Contrib/METU-NCC/MAT100/SE1/Line.pg
@@ -44,10 +44,10 @@ Context()->variables->are(x=>"Real",y=>"Real");
 $ans[0] = ImplicitPlane(Point($a,$b),Vector($d-$b,$a-$c));
 
 Context("Fraction");
-my $m = Fraction($c-$a,$d-$b); 
+my $det = $a*$d - $b*$c; 
 
-$ans[1] = Fraction($a-$c/$m); 
-$ans[2] = Fraction($c-$m*$a);
+$ans[1] = Fraction($det,$d-$b); 
+$ans[2] = Fraction($det,$a-$c);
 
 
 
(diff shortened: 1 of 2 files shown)


drdrew42:
> There is a problem when $ans[2] is a fraction, since it is getting rounded off when being plugged into $ans[1] (in the computation of $ans[3]) yielding too much rounding error for correct answers to be accepted.

It is not an issue with the rounding of $ans[2], rather it is how $ans[3] ends up being cast as a Fraction object (because of the Fraction context and the result of division when calculating $ans[3]). Because $ans[3] ends up being cast as a Fraction, then decimal answers will only be accepted when the correct answer is a terminating decimal and the submitted answer provides all decimal places. This is a byproduct of the answer checker for Fraction objects. See openwebwork/pg#429 and openwebwork/pg#430. for more details.

Explicitly casting $ans[3] (and $ans[2]) as Real objects will solve your issue. 
```perl
$ans[2] = Real( ($m*$y[0]-$q[0]) / (2*$m) );
$ans[3] = Real($ans[1]->eval(x=>$ans[2]));
```
Or, eliminating the use of the Fraction context will also resolve the issue. It is not really providing any functionality except for displaying the slope as a fraction. This can be accomplished in Numeric context by setting the context flag `reduceConstants=>0`.

benwalt:
I've modified the file as you suggested, changing the context to numeric is a more elegant solution.

I've also added another correction to a different problem... we must have graded this short exam problem by hand, and thus not noticed that webwork was generating an incorrect solution..  (#^.^#)

drdrew42 (approved):

drdrew42:
Excellent, thanks for your contributions!

PR #901, opened 2021-11-19
Fix chapter/section of probability problem (binomial, not CLT)

d-torrance:


drdrew42 (approved):
Good catch!

drdrew42:
This will require an OPL-update in order to show the problem in its new subject/chapter/section...

PR #902, opened 2021-11-23
Update Basis_30.pg

pschan-gh:
The original "correct" answer "<1, 0, 0>, <0, 1, 0>" implies that the 3-rd entry of every vector in the column space is equal to 0. We cannot know that based on the RREF of the matrix.

Also, cast a1, a2, ... as "Vector3D", otherwise an (incorrect) answer of the form <a, b, c> would throw an error message indicating wrong answer type.

### DIFF
--- a/OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
+++ b/OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
@@ -66,7 +66,7 @@ $list_checker = sub {
   foreach my $el (@$student) {
     my $parsed_el = $el;
     # Mark any answer involving explicit nonzero vectors as incorrect
-    if ($el->eval(a1=>$zero,a2=>$zero,a3=>$zero,a4=>$zero) ne $zero) {
+    if ($el->eval(a1=>$zero,a2=>$zero,a3=>$zero,a4=>$zero) != $zero) {
         return 0;
     }
     $parsed_el = $el->eval(a1=>$a1,a2=>$a2,a3=>$a3,a4=>$a4);
@@ -74,7 +74,7 @@ $list_checker = sub {
   }
 
   # return the results of the current answer checker called with the parsed array
-  return basis_checker_list_of_vectors([Vector(1, 0, 0), Vector(0, 1, 0)], $parsed_student, $ansHash, $value);
+  return basis_checker_list_of_vectors([$a1, $a2], $parsed_student, $ansHash, $value);
 };
 
 
@@ -94,17 +94,31 @@ $mc -> extra(
 );
 #$mc -> makeLast("None of the above");
 
-$ans3 = Compute("$a a1 + $c a2");
-#$ans4 = Compute("$b a1 + $d a2");
+$ans3 = Compute("$a a1 + $c a2")->cmp(
+  checker => sub{
+    my($correct,$student,$ansHash) = @_;
+    # Only answers in terms of a1 and a2 are allowed
+    foreach my $var (keys %{$student->{variables}}) {
+      Value::Error("You are not allowed to use '$var' in this answer.") if !(grep $_ eq $var,  ('a1','a2'));
+    }
+    if ($student->eval(a1=>$zero,a2=>$zero,a3=>$zero,a4=>$zero) != $zero) {
+      Value::Error("You are only allowed to use 'a1' and 'a2' in this answer.");
+    }
+    return $correct == $student;
+  }
+);
 
 $ans4 = Compute("$b a1 + $d a2")->cmp(
   checker => sub{
     my($correct,$student,$ansHash) = @_;
-    # Mark any answer involving explicit nonzero vectors as incorrect
-    if ($student->eval(a1=>$zero,a2=>$zero,a3=>$zero,a4=>$zero) ne $zero) {
-        return 0;
+    # Only answers in terms of a1, a2 and a3 are allowed
+    foreach my $var (keys %{$student->{variables}}) {
+      Value::Error("You are not allowed to use '$var' in this answer.") if !(grep $_ eq $var,  ('a1','a2','a3'));
     }
-    return $a4 == $student->substitute(a1=>$a1,a2=>$a2,a3=>$a3);
+    if ($student->eval(a1=>$zero,a2=>$zero,a3=>$zero,a4=>$zero) != $zero) {
+      Value::Error("You are only allowed to use 'a1', 'a2' and 'a3' in this answer.");
+    }
+    return $a4 == $student->eval(a1=>$a1,a2=>$a2,a3=>$a3);
   }
 );
 
--- a/OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
+++ b/OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
@@ -30,13 +30,6 @@ $showPartialCorrectAnswers = 1;
 
 Context('Matrix');
 
-Context()->variables->are(
-'a1'=>["Vector3D",'TeX'=>'\mathbf{a_1}'],
-'a2'=>["Vector3D",'TeX'=>'\mathbf{a_2}'],
-'a3'=>["Vector3D",'TeX'=>'\mathbf{a_3}'],
-'a4'=>["Vector3D",'TeX'=>'\mathbf{a_4}'],
-);
-
 do {
   ($a,$b,$c,$d) = (non_zero_random(-5,5,1),non_zero_random(-5,5,1),non_zero_random(-5,5,1),non_zero_random(-5,5,1));
 } until ($a != $b  &&  $a != $c  &&  $c != $d  &&  $b != $d);
@@ -46,6 +39,47 @@ $rrefA = Matrix([
 [0,1,$c,$d],
 [0,0,0,0]
 ]);
+
+
+$P = Matrix([[1,0,0], [0,1,0], [0, 0, 1]]);
+
+$paddedA = $P*$rrefA;
+
+Context()->variables->are(
+'a1'=>['Vector3D','TeX'=>'\mathbf{a_1}'],
+'a2'=>['Vector3D','TeX'=>'\mathbf{a_2}'],
+'a3'=>['Vector3D','TeX'=>'\mathbf{a_3}'],
+'a4'=>['Vector3D','TeX'=>'\mathbf{a_4}'],
+);
+
+
+$list_checker = sub {
+  my ($correct, $student, $ansHash, $value) = @_;
+
+  # pull the column vectors into variables
+  my ($a1,$a2,$a3,$a4) = (
+    Vector($rrefA->column(1)),
+    Vector($rrefA->column(2)),
+    Vector($rrefA->column(3)),
+    Vector($rrefA->column(4))
+  );
+  
+  # loop through the student list and parse any formulas in the student response
+  my $parsed_student = [];
+  foreach my $el (@$student) {
+    my $parsed_el = $el;
+    if ($el->class eq 'Formula') {
+      $parsed_el = $el->eval(a1=>$a1,a2=>$a2,a3=>$a3,a4=>$a4);
+    } else {
+        return 0;
+    }
+    push @$parsed_student, $parsed_el;
+  }
+
+  # return the results of the current answer checker called with the parsed array
+  return basis_checker_list_of_vectors([Vector(1, 0, 0), Vector(0, 1, 0)], $parsed_student, $ansHash, $value);
+};
+
 
 $mc = new_checkbox_multiple_choice();
 $mc -> qa (
@@ -81,6 +115,8 @@ $ans4 = Compute("$b a1 + $d a2")->cmp(
 $w1 = WordCompletion(['1','2','3','4','R','R^2','R^3','R^4'],'R^3');
 $w2 = WordCompletion(['1','2','3','4','R','R^2','R^3','R^4'],'R^4');
 
+
+
 $x1 = Vector(-1*$a, -1*$c, 1, 0);
 $x2 = Vector(-1*$b, -1*$d, 0, 1);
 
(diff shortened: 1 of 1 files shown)


dlglin:
I agree that the original answer to part e is incorrect.  This PR allows for the correct answer most likely to be entered by students, but there are other correct answers (e.g. a1, a1+a2).  If the answers are explicit vectors then this can be handled with the basis checker used elsewhere in this problem, but since the answer now depends on variables I don't think that will work.  I don't immediately see an easy way to properly check answers for this part.  It could probably be done with a custom answer checker.

I also think that the "is a subspace of ..." questions need to be fixed.  Since the student is presented with an answer blank there are many correct answers (e.g. in part d any subspace of R^3 that contains col(A)).  The student will not find out that there are a fixed number of options until they click preview or submit.  Also, why are 1,2,3,4 allowed options for the word completion?
This would be much cleaner if it were rewritten in one of the following ways:

- Replace the wordCompletions with a PopUps.
- Reword the questions to say "is a subspace of [` \mathbb{R}^n `], where [` n `]=[___]"

pschan-gh:
> I agree that the original answer to part e is incorrect. This PR allows for the correct answer most likely to be entered by students, but there are other correct answers (e.g. a1, a1+a2). If the answers are explicit vectors then this can be handled with the basis checker used elsewhere in this problem, but since the answer now depends on variables I don't think that will work. I don't immediately see an easy way to properly check answers for this part. It could probably be done with a custom answer checker.
> 

Good point. I guess one quick fix would be to use a1, a2, a3, a4 as aliases for the columns of the RREF, padded with zeros to be in R^n for some random n, and use the basis checker in the answer checker. This of course is far from ideal. I need to think more about this. Thank you!

> I also think that the "is a subspace of ..." questions need to be fixed. Since the student is presented with an answer blank there are many correct answers (e.g. in part d any subspace of R^3 that contains col(A)). The student will not find out that there are a fixed number of options until they click preview or submit. Also, why are 1,2,3,4 allowed options for the word completion? This would be much cleaner if it were rewritten in one of the following ways:
> 
> * Replace the wordCompletions with a PopUps.
> * Reword the questions to say "is a subspace of [`\mathbb{R}^n`], where [`n`]=[___]"

I like your suggestion (wordCompletions currently doesn't seem to work with MathQuill anyway ... ). However, I'm not sure if I should change this in the PR, since perhaps some instructors have been using this problem, and they actually like using wordCompletion here.

I really like this problem, and a similar one named "Basis_31.pg". But I think they both need to go through significant revision. It'd be such a shame to let them go to waste...

drdrew42:
> wordCompletions currently doesn't seem to work with MathQuill anyway

As of WeBWorK v2.16, I believe MathQuill is now automatically disabled for wordCompletion answer blanks. I've been successfully using wordCompletion with MathQuill enabled this semester.

> I really like this problem, and a similar one named "Basis_31.pg". 

Yeah, wholeheartedly agree -- and thanks to @paultpearson for these fantastic Linear Algebra problems. 

So, as far as making part _e_ more flexible, I've tested out the following 'replacement' checker:
```perl
$list_checker = sub {
  my ($correct, $student, $ansHash, $value) = @_;

  # pull the column vectors into variables
  my ($a1,$a2,$a3,$a4) = (
    Vector($rrefA->column(1)),
    Vector($rrefA->column(2)),
    Vector($rrefA->column(3)),
    Vector($rrefA->column(4))
  );

  # loop through the student list and parse any formulas in the student response
  my $parsed_student = [];
  foreach my $el (@$student) {
    my $parsed_el = $el;
    if ($el->class eq 'Formula') {
      $parsed_el = $el->eval(a1=>$a1,a2=>$a2,a3=>$a3,a4=>$a4);
    }
    push @$parsed_student, $parsed_el;
  }

  # return the results of the current answer checker called with the parsed array
  return basis_checker_list_of_vectors($correct, $parsed_student, $ansHash, $value);
};
```

I didn't write the checker to parse the 'correct' answers, so go back to using explicit vectors in the answer object for part _e_:
```perl
List(Vector(1,0,0),Vector(0,1,0))->cmp(list_checker=>$list_checker)
```

pschan-gh:
> > wordCompletions currently doesn't seem to work with MathQuill anyway
> 
> As of WeBWorK v2.16, I believe MathQuill is now automatically disabled for wordCompletion answer blanks. I've been successfully using wordCompletion with MathQuill enabled this semester.
> 

Oh, that's great, thanks!

> > I really like this problem, and a similar one named "Basis_31.pg".
> 
> Yeah, wholeheartedly agree -- and thanks to @paultpearson for these fantastic Linear Algebra problems.
> 
> So, as far as making part _e_ more flexible, I've tested out the following 'replacement' checker:
> 
> ```perl
> $list_checker = sub {
>   my ($correct, $student, $ansHash, $value) = @_;
> 
>   # pull the column vectors into variables
>   my ($a1,$a2,$a3,$a4) = (
>     Vector($rrefA->column(1)),
>     Vector($rrefA->column(2)),
>     Vector($rrefA->column(3)),
>     Vector($rrefA->column(4))
>   );
> 
>   # loop through the student list and parse any formulas in the student response
>   my $parsed_student = [];
>   foreach my $el (@$student) {
>     my $parsed_el = $el;
>     if ($el->class eq 'Formula') {
>       $parsed_el = $el->eval(a1=>$a1,a2=>$a2,a3=>$a3,a4=>$a4);
>     }
>     push @$parsed_student, $parsed_el;
>   }
> 
>   # return the results of the current answer checker called with the parsed array
>   return basis_checker_list_of_vectors($correct, $parsed_student, $ansHash, $value);
> };
> ```
> 
> I didn't write the checker to parse the 'correct' answers, so go back to using explicit vectors in the answer object for part _e_:
> 
> ```perl
> List(Vector(1,0,0),Vector(0,1,0))->cmp(list_checker=>$list_checker)
> ```

Very nice, thank you! I have incorporated this answer checker in the PR, modified to return  "incorrect" if an explicit vector <a, b, c> is entered (for part (e)).
A remaining drawback is that an answer in "mixed" form "a1 + <1, 0, 0>, a2" would still be marked correct, even though it shouldn't (since the mystery a1 might happen to be <-1, 0, 0>).

OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
```diff
@@ -102,13 +138,13 @@ c. If possible, write [` \mathbf{a_4} `] as a linear combination of [` \mathbf{a
 
 d. The dimension of the column space of [` A `] is [_____]{"2"}, and the column space of [` A `] is a subspace of [@ ANS($w1->cmp); $w1->ans_rule(5); @]*.
 
-e. Find a basis for the column space of [` A `].  Enter your answer as a comma separated list of vectors.  Each vector should have the form [| <a,b,c> |] or [| <a,b,c,d> |] where a,b,... are numbers.  Do not use the symbols [` \mathbf{a_1, a_2, \ldots} `] in your answers. 
- 
-    A basis for the column space of [`A`] is [`\big\lbrace`] [_________________________]{List(Vector(1,0,0),Vector(0,1,0))->cmp(list_checker=>~~&basis_checker_list_of_vectors)} [`\big\rbrace`]
+e. Find a basis for the column space of of [` A `].  Enter your answer as a comma separated list of vectors.  Each vector should have the form [| <a,b,c> |] or [| <a,b,c,d> |] where a,b,... are numbers.  Do not use the symbols [` \mathbf{a_1, a_2, \ldots} `] in your answers.
```
dlglin, line 141: It looks like this is introducing a typo ("of of").  Also, the instructions need to change since the correct answer requires students to enter symbols.  It's not necessary to tell students not to enter explicit vectors since the checker handles them (though they will always be marked wrong), unless this is intended as a hint that the correct answer will be in terms of a1, a2, a3, a4.

OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
```diff
@@ -102,13 +138,13 @@ c. If possible, write [` \mathbf{a_4} `] as a linear combination of [` \mathbf{a
...
- 
-    A basis for the column space of [`A`] is [`\big\lbrace`] [_________________________]{List(Vector(1,0,0),Vector(0,1,0))->cmp(list_checker=>~~&basis_checker_list_of_vectors)} [`\big\rbrace`]
+e. Find a basis for the column space of of [` A `].  Enter your answer as a comma separated list of vectors.  Each vector should have the form [| <a,b,c> |] or [| <a,b,c,d> |] where a,b,... are numbers.  Do not use the symbols [` \mathbf{a_1, a_2, \ldots} `] in your answers.   
+
+    A basis for the column space of [`A`] is [`\big\lbrace`] [_________________________]{List(a1, a2)->cmp(list_checker=>$list_checker)} [`\big\rbrace`]
 
 f. The dimension of the null space of [` A `] is [_____]{"2"}, and the null space of [` A `] is a subspace of [@ ANS($w2->cmp); $w2->ans_rule(5); @]*.
 
-g. Find a basis for the null space of [` A `].  Enter your answer as a comma separated list of vectors of the form [| <a,b,c> |] or [| <a,b,c,d> |] where a,b,... are numbers.
+g. Find a basis for the null space of of [` A `].  Enter your answer as a comma separated list of vectors of the form [| <a,b,c> |] or [| <a,b,c,d> |] where a,b,... are numbers.
```
dlglin, line 147: Another typo ("of of")

dlglin (changes requested):
The typos and instructions mentioned below should be fixed before merging.

My vote would be to change the subspace questions to PopUps.  Since there are other correct answers that could otherwise be typed in the answer blank, I think having an answer blank is misleading.  I also think that it is a bad experience for students to be able to type answers that aren't allowed, and only later be told that their answer isn't accepted.

OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
```diff
@@ -47,6 +40,47 @@ $rrefA = Matrix([
 [0,0,0,0]
 ]);
 
+
+$P = Matrix([[1,0,0], [0,1,0], [0, 0, 1]]);
+
+$paddedA = $P*$rrefA;
```
drdrew42, line 46: I don't see these being used, they should be deleted.

drdrew42 (changes requested):
Is it really desirable to require that students input `a1`, `a2` (or combinations thereof), over explicit vectors? Why not allow students to do either?

OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
```diff
@@ -47,6 +40,47 @@ $rrefA = Matrix([
...
+  
+  # loop through the student list and parse any formulas in the student response
+  my $parsed_student = [];
+  foreach my $el (@$student) {
+    my $parsed_el = $el;
+    if ($el->class eq 'Formula') {
+      $parsed_el = $el->eval(a1=>$a1,a2=>$a2,a3=>$a3,a4=>$a4);
+    } else {
+        return 0;
+    }
```
drdrew42, line 75: So students are _required_ to list formulas then? The instructions still say, "Each vector should have the form [| <a,b,c> |] or [| <a,b,c,d> |] where a,b,... are numbers" but this is no longer true.

OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
pschan-gh: Oh no. My bad, I somehow cut-and-pasted an earlier version of the question back. Yes, you are right, it's not necessary to tell students not to enter explicit vectors.

OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
pschan-gh: Indeed, the requirement that explicit vectors be entered should be removed. My bad.

The answer checker will still accept explicit vectors. It's just that they'd be marked incorrect.

OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
pschan-gh: > This works for me. Based on my testing it accepts all and only the correct answers.
> 
> The typos and instructions mentioned below should be fixed before merging.
> 
> My vote would be to change the subspace questions to PopUps. Since there are other correct answers that could otherwise be typed in the answer blank, I think having an answer blank is misleading. I also think that it is a bad experience for students to be able to type answers that aren't allowed, and only later be told that their answer isn't accepted.

I agree. It seems wordCompletion is useful for problems that involve longer proof sentences, or many choices. Using it for one-word answers seems a bit of an overkill. But I am neutral on this. Does anyone else have any opinion on this?

pschan-gh: Yes, thanks!

pschan-gh:
PR has been updated based on the comments of @dlglin and @drdrew42 (Thanks a lot!).

The latest update changed the WordCompletion questions to Pop-Ups. I agree with @dlglin's view on this, but I am also open to any objection to the change.

OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
```diff
@@ -95,7 +95,18 @@ $mc -> extra(
 #$mc -> makeLast("None of the above");
 
 $ans3 = Compute("$a a1 + $c a2");
-$ans4 = Compute("$b a1 + $d a2");
+#$ans4 = Compute("$b a1 + $d a2");
+
```
drdrew42: No need to keep this hanging around.

OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
```diff
@@ -95,7 +95,18 @@ $mc -> extra(
...
+$ans4 = Compute("$b a1 + $d a2")->cmp(
+  checker => sub{
+    my($correct,$student,$ansHash) = @_;
+    # Mark any answer involving explicit nonzero vectors as incorrect
+    if ($student->eval(a1=>$zero,a2=>$zero,a3=>$zero,a4=>$zero) ne $zero) {
+        return 0;
+    }
+    return $a4 == $student->eval(a1=>$a1,a2=>$a2,a3=>$a3, a4=>$zero);
+  }
+);
```
drdrew42: This checker should include the following:
```perl
    # a4 should not be allowed in the student formula
    foreach my $var (keys %{$student->{variables}}) {
      Value::Error('You are not allowed to use "a4" in this answer') if $var eq 'a4';
    }
```

drdrew42: and I would also recommend using `Value::Error()` to provide feedback in the case where a student answer uses explicit vectors...

drdrew42 (changes requested):
answer checker for part 4 needs revision. Students can submit a combination of a1, a2, a3 that adds up to a4, then add any multiple of a4 to that answer and still receive credit since a4 is zeroed out.

OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
drdrew42: and L104 should use `==` to test equality with the zero vector, and not `eq` (that's for comparing strings)

drdrew42: I see that you changed eval to substitute so that you could drop the `a4` -- but I think it's better to recognize and provide feedback when `a4` is used in an answer.

pschan-gh: >  and I would also recommend using Value::Error() to provide feedback in the case where a student answer uses explicit vectors...

Thanks for the Value::Error tip. I am wondering what's an appropriate way to detect explicit vectors. Zeroing out the variables, and then comparing with <0,0,0>, still accepts an answer with <0,0,0> in it, which technically should not be allowed in Parts (b) and (c). I guess I could use regex ($student !~ /<|[ijk]/), but perhaps there are better ways? Thanks!

pschan-gh: Pardon the hiatus. A new update has been made to include error messages for using "forbidden" vectors.

OpenProblemLibrary/Hope/Multi1/03-05-Basis-subspace/Basis_30.pg
```diff
@@ -63,23 +94,39 @@ $mc -> extra(
...
-#$ans4 = Compute("$b a1 + $d a2");
+$ans3 = Compute("$a a1 + $c a2")->cmp(
+  checker => sub{
+    my($correct,$student,$ansHash) = @_;
+    # Only answers in terms of a1 and a2 are allowed
+    foreach my $var (keys %{$student->{variables}}) {
+      Value::Error("You are not allowed to use '$var' in this answer.") if !(grep $_ eq $var,  ('a1','a2'));
+    }
+    if ($student->eval(a1=>$zero,a2=>$zero,a3=>$zero,a4=>$zero) != $zero) {
+      Value::Error("You are only allowed to use 'a1' and 'a2' in this answer.");
```
drdrew42, line 105:
```suggestion
      Value::Error("You are only allowed to use linear combinations of 'a1' and 'a2' in this answer.");
```

PR #907, opened 2021-12-16
fixed rounding issues

brittnilorton:
some rounding fixes, a domain issue fix, and a video added

### DIFF
--- a/Contrib/CCCS/AlgebraicLiteracy/IA_9.5/OpenStax_IA_9.5_203.pg
+++ b/Contrib/CCCS/AlgebraicLiteracy/IA_9.5/OpenStax_IA_9.5_203.pg
@@ -66,7 +66,7 @@ tolerance => 0.005,
 BEGIN_PGML
 The area of a triangular flower bed in the park is [`[$A]`] square feet.  The base is [`[$m]`] feet longer than twice the height.  Find the base and height.  Do not round until the end, then round to the nearest tenth of a foot.
 
-Height = [_______________]{$height}feet
+Height = [_______________]{$heightr}feet
 
 Base =  [_______________]{$baser}feet
 
--- a/Contrib/CCCS/AlgebraicLiteracy/IA_9.5/OpenStax_IA_9.5_203.pg
+++ b/Contrib/CCCS/AlgebraicLiteracy/IA_9.5/OpenStax_IA_9.5_203.pg
@@ -49,16 +49,26 @@ $a2 = -$c;
 $height = Compute("(-$b + sqrt($b**2 - 4*$a*$c))/(2*$a)");
 $base = Compute("2*$height + $m");
 
+$heightr = Round($height,1);
+$heightr = Compute ($heightr)->with(
+tolType => 'absolute',
+tolerance => 0.005,
+);
 
+$baser = Round($base,1);
+$baser = Compute ($baser)->with(
+tolType => 'absolute',
+tolerance => 0.005,
+);
 ###########################
 #  Main text
 
 BEGIN_PGML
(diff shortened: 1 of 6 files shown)


Contrib/CCCS/AlgebraicLiteracy/IA_9.5/OpenStax_IA_9.5_203.pg
```diff
@@ -49,16 +49,26 @@ $a2 = -$c;
...
 #  Main text
 
 BEGIN_PGML
-The area of a triangular flower bed in the park is [`[$A]`] square feet.  The base is [`[$m]`] feet longer than twice the height.  Find the base and height.  Round to the nearest tenth.
+The area of a triangular flower bed in the park is [`[$A]`] square feet.  The base is [`[$m]`] feet longer than twice the height.  Find the base and height.  Do not round until the end, then round to the nearest tenth of a foot.
 
 Height = [_______________]{$height}feet
 
-Base =  [_______________]{$base}feet
+Base =  [_______________]{$baser}feet
```
drdrew42, line 71: Did you intend to change `$height` to `$heightr` here as well?

Contrib/CCCS/AlgebraicLiteracy/IA_9.5/OpenStax_IA_9.5_216.pg
```diff
@@ -33,15 +33,16 @@ $showPartialCorrectAnswers = 1;
...
 
 $v0 = random(150, 300, 10);
 $h0 = random(3, 6);
-$t = Compute("(-$v0 - sqrt($v0^2 + 64*$h0))/-32");
+
+$t = Compute("(-$v0 - sqrt($v0**2 + 64*$h0))/-32");
+$t = Round($t,2);
+$t = Compute ($t)->with(
+tolType => 'absolute',
+tolerance => 0.0005,
```
drdrew42, line 44: This will require accuracy to the thousandth of a second.

drdrew42 (changes requested):
See comments for recommended changes.

Contrib/CCCS/AlgebraicLiteracy/IA_9.5/OpenStax_IA_9.5_203.pg
brittnilorton: Yes, yes we did. I will correct.

Contrib/CCCS/AlgebraicLiteracy/IA_9.5/OpenStax_IA_9.5_216.pg
brittnilorton: I was under the impression this was a 'better safe than sorry' thing because this is the tolerance and the rounding is in the line above. But I will correct this to have tolerance = > 0.005,

drdrew42: I see -- so you're enforcing that the rounding took place?

brittnilorton: Correct. Ensuring that any other rounding would be marked incorrect, although perhaps this is not necessary.

drdrew42: No, probably not necessary -- but it does enforce the instructions in the problem. Since that is what you intended, there's no need to change it (in a way that would accept answers that are not rounded according to the instructions).

PR #909, opened 2022-01-19
SFU SEE 110 and SEE 310 problems

trevorb1:
This pull request creates the new folder `Contrib/SFU/` to hold WeBWork problems for Simon Fraser University. Two folders are added here for the Sustainable Energy Engineering classes, `SEE110/` and `SEE310/`. All problems have been tagged, tested on at least one class of students (~40 people), and include full solutions. We are hoping that all submitted problems will be considered for the Open Problem Library! 

The problems submitted primarily focus on energy literacy and energy system modelling concepts. When we were reviewing the existing problems on the OPL, we were unable to find any problem sections that matched these concepts. As such, we created our own problems and followed the tagging system shown below: 

- Sustainable Energy Engineering (DBsubject)
  - Energy Conversion (DBchapter)
    - Unit Conversion (DBsection)
    - Applications
  - Energy System
    - Energy Demand
    - Energy Supply
    - Applications
    - Energy Chains
    - Storage
  - Energy System Modelling
    - Input Output Models 
    - Linear Programming Models

The Energy Conversion and Energy System chapters include concepts such as converting between electrical energy and power, calculating emissions and carbon tax, finding material use, calculating levelized cost of electricity, ect… The closest existing questions we could find to these approached the problems from a mechanical/thermodynamic perspective, rather then an energy system perspective. The last chapter (Energy System Modelling) covers two types of energy modelling techniques. We saw that the operations research chapter has linear programming fundamental questions, however, our lp problems are more application based rather then theory based.  

We would be happy to discuss and implement any feedback you have on the pr. Thank you in advance!


drdrew42 (approved):
Thanks so much for your contribution -- the DB chapter/subject/section taxonomy is in the process of expanding to cover more fields... thanks for your input!

PR #910, opened 2022-01-26
Add AnswerFormatHelp

taniwallach:
Add use of AnswerFormatHelp to 3 problems (two on vectors, one on points).
Since the page at https://webwork.maa.org/wiki/Available_Functions  does not mention vectors or points, having these links is quite helpful if these are problems which require such syntax and are early in the course.

One problem has a line `DEBUG` at the top of the question, and it seems that is an accidental leftover from https://github.com/taniwallach/webwork-open-problem-library/commit/32a3f8100123b7417d62376f631c2cc4ce4a69df#diff-123982bdc1fa0ddfa9dc8e7c03c6b44e20947d868d5db9c2bd7ce3c7bba30b75


drdrew42 (approved):
Thanks @taniwallach -- good catch on that lingering "DEBUG"

PR #914, opened 2022-02-10
Fix indexing in case R1 <-> R2

lahvak:
When swapping row 1 and row 2, there was $a12 instead of $a21.  That
would cause about 1/3 of the generated problems to have one of the 27
answers wrong.


drdrew42 (approved):
Confirmed, though it took a while to find a seed that swapped R1 and R2...

Good catch!

PR #917, opened 2022-02-22
Move bizarro arithmetics settings

lahvak:
The bizarro arithmetics was initialized before the calculations for the
problem statement were done, leading to the problem statement being
scrambled.  This commit moves them so they only apply to the answer.

### DIFF
--- a/OpenProblemLibrary/CollegeOfIdaho/setAlgebra_06_02_AddSubRationalExpressions/62IntAlg_19_AddSubRatExp.pg
+++ b/OpenProblemLibrary/CollegeOfIdaho/setAlgebra_06_02_AddSubRationalExpressions/62IntAlg_19_AddSubRatExp.pg
@@ -33,18 +33,6 @@ TEXT(beginproblem());
 
 Context("Numeric");
 
-Context()->operators->set(
-   '/' => {class => 'bizarro::BOP::divide', isCommand => 1},
-   '/ ' => {class => 'bizarro::BOP::divide', isCommand => 1},
-   ' /' => {class => 'bizarro::BOP::divide', isCommand => 1},
-   '//' => {class => 'bizarro::BOP::divide', isCommand => 1},
-   '*' => {class => 'bizarro::BOP::multiply', isCommand => 1},
-   '* ' => {class => 'bizarro::BOP::multiply', isCommand => 1},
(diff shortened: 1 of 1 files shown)


lahvak:
Note that there seems to be a relevant discussion at https://webwork.maa.org/moodle/mod/forum/discuss.php?d=4948, however, I don't seem to be able to reset my password in order to contribute there.

drdrew42:
Looking at the linked issue from the forums, and doing some investigating of my own, it seems that there is no _spaced_ version of addition in PG. The spaced versions of multiplication and division allow for certain distinctions, but no such distinction is supported for addition. The correct fix here is to simply remove the two lines declaring '+ ' and ' +' operators, they have the potential to be problematic no matter _where_ they are declared in the code.

drdrew42:
You can find the default operators in `lib/Parser/Context/Default.pm` in the openwebwork/pg repo, lines 39-108.

OpenProblemLibrary/CollegeOfIdaho/setAlgebra_06_02_AddSubRationalExpressions/62IntAlg_19_AddSubRatExp.pg
```diff
@@ -64,6 +52,18 @@ $den2 = Formula("y + $d")->reduce;
...
+   '/' => {class => 'bizarro::BOP::divide', isCommand => 1},
+   '/ ' => {class => 'bizarro::BOP::divide', isCommand => 1},
+   ' /' => {class => 'bizarro::BOP::divide', isCommand => 1},
+   '//' => {class => 'bizarro::BOP::divide', isCommand => 1},
+   '*' => {class => 'bizarro::BOP::multiply', isCommand => 1},
+   '* ' => {class => 'bizarro::BOP::multiply', isCommand => 1},
+   ' *' => {class => 'bizarro::BOP::multiply', isCommand => 1},
+   '+' => {class => 'bizarro::BOP::add', isCommand => 1},
+   '+ ' => {class => 'bizarro::BOP::add', isCommand => 1},
+   ' +' => {class => 'bizarro::BOP::add', isCommand => 1},);
```
drdrew42, line 65:
```suggestion
```

drdrew42 (changes requested):

drdrew42 (approved):

PR #918, opened 2022-02-25
display fixes

brittnilorton:
a couple of display fixes and directions updated

### DIFF
--- a/Contrib/CCCS/PreCalculus/3.2/RRCC_CCCS_Openstax_AlgTrig_AT-1-001-AS_3_2_43.pg
+++ b/Contrib/CCCS/PreCalculus/3.2/RRCC_CCCS_Openstax_AlgTrig_AT-1-001-AS_3_2_43.pg
@@ -53,8 +53,6 @@ add_functions($graph, "$slope1*(x-$x1)+$y1 for x in <$x1,$x2>".
 add_functions($graph, "$slope2*(x-$x2)+$y2 for x in <$x2,$x3>". 
   " using color:blue and weight:3");
 $graph->stamps(open_circle($x1,$y1,'blue'), open_circle($x2,$y2,'blue'), open_circle($x3, $y3,'blue')  );
-$in=time();
-$graph->gifName($graph->gifName()."$in");
 
 $piecewise =
        BeginTable().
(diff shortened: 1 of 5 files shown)


Contrib/CCCS/PreCalculus/3.2/RRCC_CCCS_Openstax_AlgTrig_AT-1-001-AS_3_2_43.pg
```diff
@@ -33,41 +39,49 @@ $x3 = random(3,5,1);
...
 $graph = init_graph(-6,-6,6,6,grid =>[12,12], axes => [0,0]);
-$f1 = new Fun( plot_list([$x1, $x2, $x3],[ $y1,$y2,$y3]),  $graph);
+#$f1 = new Fun( plot_list([$x1, $x2, $x3],[ $y1,$y2,$y3]),  $graph);
+add_functions($graph, "$slope1*(x-$x1)+$y1 for x in <$x1,$x2>". 
+  " using color:blue and weight:3");
+add_functions($graph, "$slope2*(x-$x2)+$y2 for x in <$x2,$x3>". 
+  " using color:blue and weight:3");
 $graph->stamps(open_circle($x1,$y1,'blue'), open_circle($x2,$y2,'blue'), open_circle($x3, $y3,'blue')  );
+$in=time();
+$graph->gifName($graph->gifName()."$in");
```
drdrew42, line 57: This is going to cause each student to generate a new image file with every attempt, creating unnecessary bloat in the course's `tmp/` folder. 

What is this aiming to accomplish? (There's probably a better way...)

brittnilorton:
Oh! I didn't realize that. This piece is something we have used multiple places because I believe we thought it was a way to refresh the image rather than using the cached image. The main reason for that was while we are coding and testing different graphs, this helps the coder see the new image coded rather than a cached image. Perhaps unnecessary now...

drdrew42:
I am all too familiar with browser caching issues... but the fix should not be implemented through WeBWorK. Rather, open the developer tools in your browser and under the network tab of the developer tools you should see an option to 'disable cache':

Screen Shot 2022-02-25 at 3 58 20 PM

Note that this is only effective while the developer tools are open.

Definitely look back through your catalog of problems for other places where this has been implemented, it is code that you'll certainly want to remove.

brittnilorton:
Gotcha. I'll be sure to let my team know of that for any future problems they code. I'll spend the time this semester removing those lines from our existing problems. Thanks!

drdrew42 (approved):
👍

PR #919, opened 2022-03-01
Define missing pg variables

lahvak:
Variables $d and $D were used in the text but never defined.


drdrew42:
Good catch! However, I still get errors rendering this problem. The "NO" string is declared in all caps, but is aliased by "N" and "NONE" as _'no'_. Adding `caseSensitive=>0` only applies to student responses, so the best resolution IMO is changing the declaration of "NO" to lowercase.

PR #920, opened 2022-03-01
Rewrite problem with MathObjects,PGML in order to disallow use of sqrt

dlglin:
Other than making it more current, this PR makes three significant changes to the problem:

1. Students can no longer use the sqrt function in the answer (or exponentiation), which would allow them to get it correct without any work.
2. The correct answer now always has nice values, which was not the case previously.
3. The question now asks for both square roots, rather than just the one with positive Real component.

2 and 3 could be considered modifications to the spirit of the problem rather than bug fixes, but I consider them both improvements.  If there are objections to replacing the problem then I can resubmit as a new problem.

### DIFF
--- a/OpenProblemLibrary/Rochester/setAlgebra11ComplexNumbers/Sqrt.pg
+++ b/OpenProblemLibrary/Rochester/setAlgebra11ComplexNumbers/Sqrt.pg
@@ -2,13 +2,12 @@
 ##   Complex Numbers
 ## ENDDESCRIPTION
 
-## Tagged by nhamblet
-
 ## DBsubject(Complex analysis)
 ## DBchapter(Arithmetic)
 ## DBsection(Powers and roots)
 ## Institution(Rochester)
 ## Level(2)
+## MO(1)
 ## KEYWORDS('Complex', 'Imaginary')
 
 DOCUMENT();        # This should be the first executable line in the problem.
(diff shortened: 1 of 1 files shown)


OpenProblemLibrary/Rochester/setAlgebra11ComplexNumbers/Sqrt.pg
```diff
@@ -2,43 +2,41 @@
...
+  "PGML.pl",
   "PGcourse.pl"
 );
 
 TEXT(beginproblem());
 $showPartialCorrectAnswers = 1;
 
+Context("Complex");
+Context()->functions->disable('sqrt');
+Context()->operators->undefine('^');
```
drdrew42, line 28:
```suggestion
Context()->operators->undefine('^', '**');
Context()->{error}{msg}{"Can't use '^' in this context"} 
  = "Exponents are disabled for this problem. Compute the result for yourself and resubmit your answer.";
Context()->{error}{msg}{"Can't use '**' in this context"} 
  = "Exponents are disabled for this problem. Compute the result for yourself and resubmit your answer.";
Context()->{error}{msg}{"Function 'sqrt' is not allowed in this context"} 
  = "Square roots are disabled for this problem. Compute the result for yourself and resubmit your answer.";
```

drdrew42: Otherwise, `**` can still be used to give a 1/2 exponent...

drdrew42 (approved):

PR #921, opened 2022-03-03
Actually apply changes to limits and test_at.

lahvak:
The changes to limits and test_at were never applied, so the problems
sometimes fails to generate enough points (e.g. seed = 4857)

### DIFF
--- a/OpenProblemLibrary/AlfredUniv/anton8e/chapter1/1.5/prob9.pg
+++ b/OpenProblemLibrary/AlfredUniv/anton8e/chapter1/1.5/prob9.pg
@@ -44,7 +44,7 @@ TEXT(beginproblem());
 
 $ans1 = Formula("1/sqrt(x**2 + 1)")->reduce;
 $ans2 = Formula("sqrt(1-1/x**2)")->reduce;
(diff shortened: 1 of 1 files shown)


OpenProblemLibrary/AlfredUniv/anton8e/chapter1/1.5/prob9.pg
```diff
@@ -44,7 +44,7 @@ TEXT(beginproblem());
 
 $ans1 = Formula("1/sqrt(x**2 + 1)")->reduce;
 $ans2 = Formula("sqrt(1-1/x**2)")->reduce;
-$ans2->with(limits=>[1.1, 10], test_at=>[[-2],[-3]]);
+$ans2 = $ans2->with(limits=>[1.1, 10], test_at=>[[-2],[-3]]);
```
dlglin, line 47:
```suggestion
$ans2->{limits}=[1.1, 10];
$ans2->{test_at}=[[-2],[-3]];
```
I believe that this is the preferred syntax.

This problem also needs
```suggestion
Context()->functions->disable('Trig');
```
Otherwise a student can simply enter cos(arctan(x)) and get credit for the problem.

PR #924, opened 2022-04-11
Create a workflow that generates an OPL META release.

drgrice1:
This adds a workflow that will be run anytime that a tag starting with the letter "v" is pushed to openwebwork/webwork-open-problem-library.  The workflow will execute modified versions of the OPL-update and generate-OPL-set-def-lists.pl scripts from openwebwork/webwork2 bin and generate a database dump and json files that can be used by the webwork2 docker build (or directly for anyone that doesn't want to wait for OPL-update).

A corresponding pull request to openwebwork/webwork2 sets this up for docker, and adds a script that others can use to directly set up a server to use the latest release created by this workflow.

You can test this workflow by copying the branch for this pull request to your fork, and then pushing a tag that starts with the letter "v".

### DIFF
--- a/.github/workflows/bin/Tags.pm
+++ b/.github/workflows/bin/Tags.pm
@@ -1,0 +1,226 @@
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright &copy; 2000-2022 The WeBWorK Project, https://github.com/openwebwork
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of either: (a) the GNU General Public License as published by the
+# Free Software Foundation; either version 2, or (at your option) any later
+# version, or (b) the "Artistic License" which comes with this package.
+#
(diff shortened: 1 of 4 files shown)


pstaabp (approved):
This works as advertised.  @drgrice1 I see no reason to not merge this.

drgrice1:
There is one thing to decide in addition to merging this.  That is what to use for the version naming scheme.  This pull request leaves that open.  The only restriction is that it begin with the letter v.  I was thinking something like "v2022-04-14".

pstaabp:
> The only restriction is that it begin with the letter v. I was thinking something like "v2022-04-14".

I think that seems reasonable with the OPL since we won't have major/minor releases like software, which is the other standard way.

drdrew42 (approved):
I like the suggestion to name version releases as `vYYYY-MM-DD`

PR #930, opened 2022-05-03
Create SUNYOneonta folder in Contrib folder and add five folders of new problems

LBridgers:
We have recently been creating problems at SUNY Oneonta specifically for Math for El Ed/ Math for the Liberal Arts. This first pull request is the first step in getting these added to the OPL. I believe that all of the problems being added are new and different from things I have seen in the OPL. In this pull request, I am adding problems on Area and Perimeter, Surface Area and Volume, Transformations, Numeration Systems (specifically history of numeration systems), and also problems on factorization, gcd/lcm prime factorization and divisibility tests. All of these problems are randomized, and the geometry problems also include diagrams/pictures that are coded in.

I hope that all can be considered for inclusion in the OPL. We have tagged them so, I believe they are ready to be included once they have been vetted and approved.


drdrew42 (approved):
Thanks for this contribution! Before we can consider these problems for the OPL, they need to be tagged according to the existing OPL taxonomy (e.g. "Shapes" is not an official DBSubject). I'm going to go ahead and merge these problems into Contrib, and if/when you re-tag these problems, open another pull requests, and then they'll be searchable in the Problem Library.

[Official taxonomy](https://raw.githubusercontent.com/openwebwork/webwork-open-problem-library/main/OpenProblemLibrary/Taxonomy2)

The taxonomy linked above is a little tricky to interpret, but indentions indicate sub-categories starting from DBSubject -> DBChapter -> DBSection. Any DBSection with `>>>` indicates that it overlaps with another DBSection elsewhere in the taxonomy (e.g. Precalculus is made up entirely of overlapping content, mainly from Algebra & Trigonometry).

PR #932, opened 2022-05-26
Fix display of percent signs for hardcopies in 2 statistics problems

d-torrance:
% -> $PERCENT


drdrew42 (approved):
Thanks for catching this!

PR #937, opened 2022-07-14
problem fixes for OPLv3 script

pstaabp:
This fixes a number of problems found when running the OPLv3 import script.  This includes:
* deleting empty files
* files in `Contrib/Rochester-SigPrep` that had bad characters in the problem source, but  clearly are just copies.  These now point to the original file. 
* In DouglasCollege problems, replaced `[REFER TO https://github.com/openwebwork/webwork-open-problem-library/blob/master/OpenProblemLibrary/Taxonomy]` with `UNKNOWN`.
* A number of places, a non-standard single quote was replaced with a standard one.
* In one case, the problem pointed to was incorrect.


drdrew42 (approved):

drdrew42:
🙏 🙏 🙏  @pstaabp 🙏 🙏 🙏

PR #938, opened 2022-07-14
Creation of a randomPerson macro that handles pronouns other that s/he

pstaabp:
The idea of this was the `PCC/PCCmacros.pl` subroutine `RandomName` and extend to handling pronouns and the ability to pass in the correct verb conjugation for problems. 

Note: there is a test file with this.  Generally test files should not be in the same directory, but we need to find a more general way of handling these.


pstaabp:
This doesn't run in the wild.  No need to look at this yet.

pstaabp:
This is working now.  For testing,  is a sample file 

[simple_example.pg.txt](https://github.com/openwebwork/webwork-open-problem-library/files/9122229/simple_example.pg.txt)
[simple_example.pg.txt](https://github.com/openwebwork/webwork-open-problem-library/files/9122229/simple_example.pg.txt)

pstaabp:
Since all macros have been moved to PG, this is closed and moved to openwebwork/pg#761.

PR #941, opened 2022-08-05
Add the ability to include alt tags for unionImages and imageChoice macros

pstaabp:
This allows for adding extra html tags to unionImages and in imageChoice macros.  This is mainly for use for alt tags in images. 

A sample problem with images have been attached. The important part of this test is that the alt tag of the images have been set correctly--the order is important.

Note: there are two copies of `imageChoice.pl` in this repository. The one in `CollegeOfIdaho` is loaded before the one in Union, so this one is update. 

In a future PR, consolidation of the macros will be done. 

[image_choice.pg.txt](https://github.com/openwebwork/webwork-open-problem-library/files/9268012/image_choice.pg.txt)
image1
image2
image3
image4
image5
image6

### DIFF
--- a/OpenProblemLibrary/macros/CollegeOfIdaho/imageChoice.pl
+++ b/OpenProblemLibrary/macros/CollegeOfIdaho/imageChoice.pl
@@ -1,5 +1,125 @@
+################################################################################
+# WeBWorK Online Homework Delivery System
+# Copyright &copy; 2000-2022 The WeBWorK Project, https://github.com/openwebwork
+#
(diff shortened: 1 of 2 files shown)


drdrew42:
Did these changes make it into the versions of these macros when they moved over to PG? Should this get merged or is it redundant?

pstaabp:
We should probably close this.  I'm rethinking this macro anyway.  It seems like we can handle this with the parserRadioButton automatically.

pstaabp:
Closing this, since macros have been moved to PG.

PR #942, opened 2022-08-11
Switch one set's images from GIF to PNG 

mitchkeller:
This is motivated by the fact that three of these problems, all using `right_triangle.gif`, are used in Active Prelude to Calculus. We use `xelatex` to build the PDF version, and `xelatex` does not like GIF images. I left the original GIF images behind, but all the problems in this set were converted over to use the PNG version rather than the GIF version. I can push another commit that `git rm`s the GIF images if that would be preferred.


drdrew42:
This is fine, I would ask that you also commit the deletion of the (now) unused .gif files as well. Thanks!

mitchkeller:
Pushed commit that deletes the now unused .gif files.

drdrew42 (approved):
Thanks for the modifications!

PR #945, opened 2022-08-17
Minor bug fixes in macro files (reduces log messages)

paulvojta:


drdrew42:
Thanks Paul -- these warnings are so prevalent that I've added `no warnings 'redefine';` in `PG::IO` and `PG::Translator` to relieve some of the log-bloat. (Though that certainly doesn't help me catch problems that generate these warnings!)

OpenProblemLibrary/macros/Alfred/Alfredmacros.pl
```diff
@@ -27,20 +27,22 @@ sub inversetrig
...
+#};
+#return $maximum;
+#}
+#sub min {
+#my $minimum = shift @_;
+#foreach(@_){
+#   if ($minimum > $_){$minimum =  $_};
+#};
+#return $minimum;
+#}
```
drdrew42, line 45:
```suggestion
```

drdrew42: Simply delete these lines. WeBWorK has a bad habit of retaining unnecessary comments... one can always review the git history to retrieve old code, if it's ever relevant.

drdrew42 (approved):

PR #949, opened 2022-09-19
Rearrange the OPL macros for clarity

pstaabp:
This is the OPL version of the rearranging of the macros.  

This branch just copies site-specific macros to a `macros/univ` directory.   In order to test, see openwebwork/PG#724 and openwebwork/webwork2#1803

Still todo is to remove/move all macros not in the `macros/univ` directory.  Some of these can be relocated to a Contrib/macros directory and many should be delete either due to obsolescence or because they have been moved to PG.


drgrice1 (approved):
This will be needed to accommodate the webwork2 and pg pull requests.

drdrew42 (approved):
DELETED!

PR #956, opened 2022-09-26
ur_ab_7_4.pg bugfix

pschan-gh:
There was a slim chance that the random variables $a and $b collide twice in a row. (With enough students, there is no chance too slim.)


drdrew42 (approved):
A better way to restrict randomizations (especially in the case of avoiding collisions) is to use a do-until loop on the random call. This preserves the random 'range' of `1..9` and does not force the two values to be sequential whenever there is a collision.

drdrew42:
Good catch, and thanks for the fix!

PR #957, opened 2022-09-27
Bugfix: The origin is always a source, not a spiral source,

jirilebl:
The origin is always a source, not a spiral source, thanks to Xu-Yan Chen of Georgia Tech for reporting it.


jirilebl:
Actually come to think of it, while this fixes the problem there may be a better fix, let me think about it for a bit, so don't apply it yet.

jirilebl:
OK, it's fine.  No need to make it more complicated.

Feel free to apply it, sorry for the back and forth. :)

PR #967, opened 2022-10-17
Rewrite problem with PGML and convert text entry to popup menus

dlglin:


gajennings:
Danny, your version is definitely better.  I'll merge it, then I'll correct the spelling "lexicograpic" (it's wrong in the original version too). Also I'll change "critical point with the next smallest x-coordinate" to "next critical point" because of the lexicographic order.  Then I'll put those fixes in a separate pull request, and merge them too.  (Maybe there's a way to fix all these things in one merge but I don't understand github well enough to do that. Honestly, Github is a mystery to me!)

PR #973, opened 2022-10-27
Fix bug 4715

gajennings:


OpenProblemLibrary/ASU-topics/setUnitCircle/srw5_2_13.pg
```diff
@@ -8,32 +8,38 @@
 ## DBchapter(Trigonometric functions)
 ## DBsection(Trigonometric functions of special angles)
 ## Institution(ASU)
-## Level(2)
+## MO(1) ## Level(2)
```
drgrice1, line 11: @gajennings:  This change in this pull request causes the OPL-update script to fail as is an invalid tag.

drdrew42: Fixed. Thanks for catching that...

PR #1017, opened 2023-02-14
Fix custom grader to be compatible with gateway quiz mode

dlglin:
There are a number of problems in the OPL which have custom graders that are hard-coded to look for the keys `'AnSwEr0001'`, `'AnSwEr0002'`, etc. in the `%evaluated_answers` hash.  This means that they don't work in gateway sets because of the prepended gateway prefix.

This is a quick fix to one such problem.  Let me know if you think that this is the right approach.  If so, I will propagate it to the other offending problems.


drdrew42 (approved):
making a note of these file paths... they really ought to use `custom_problem_grader_fluid` instead of duplicating the behavior in the problem code. Thanks Danny!

PR #1018, opened 2023-02-17
Display Fixes and Typos

brittnilorton:
Some display issues and typos. 
One deletion (the content of that problem was merged into another, so that problem has no context anymore).


drdrew42 (approved):
These look good. Only one comment regarding upcoming changes to AnswerFormatHelp -- in the next release of WeBWorK (technically the next release of PG), it is no longer necessary to load the AnswerFormatHelp.pl macro, and authors should switch to using `helpLink()` instead. The syntax is the same, the same list of answer types is still supported, just shorter to write and one fewer macro to load!

AnswerFormatHelp isn't going to be removed or anything, just wanted to give you a heads up on how new problems can be written more simply...

PR #1019, opened 2023-02-23
Fix axis of revolution for surface area integral problem

d-torrance:
The solution to this problem assumes that we use f(x) as the radius of each slice in our surface of revolution, but this gives the distance to the x-axis, not the y-axis as stated in the problem description.


drdrew42 (approved):
Good catch!

PR #1035, opened 2023-04-03
Fix accepting answers due to excessive rounding in the

jirilebl:
Fix accepting answers due to excessive rounding in the problem setup.

First, make WW not reduce answers, and give the rational expressions exactly (simplified).  This also gives the "correct answer" to the student in a more reasonable way I think.

Also change the limits to [-1,0] as there the exponential is of reasonable size.

This should fix the previous problem that I suppose wasn't quite fixed.

This current issue was reported against a problem based on KJ-3-10-09.pg by Andras Balogh (UTRGV) for seed 4407.  That is this is https://github.com/jirilebl/diffyqs-webwork/blob/master/sec_2.6/prob01.pg

### DIFF
--- a/OpenProblemLibrary/FortLewis/DiffEq/2-Higher-order/06-Forcing-resonance/KJ-3-10-08.pg
+++ b/OpenProblemLibrary/FortLewis/DiffEq/2-Higher-order/06-Forcing-resonance/KJ-3-10-08.pg
@@ -59,10 +59,6 @@ $w = 8;
 
 $F = Formula("$a e^(-t)")->reduce();
 
-## this does too much rounding for some seeds
-#$c1 = Compute("-$a/74");
-#$c2 = Compute("($a/74 + 2*$c1)/6");
-#$c3 = Compute("$a/74");
 $c1 = Formula("-$a/74");
 $c2 = Formula("-$a/444");
 $c3 = Formula("$a/74");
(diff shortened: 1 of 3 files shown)


jirilebl:
By "the previous problem" I mean the fix that was made to KJ-3-10-09.pg in a commit long time ago by someone, where the range was changed and the tolerance was changed to absolute.  And by it not working, I mean that it still then fails for seed 4407.  I just realized I wasn't really being clear before.

OpenProblemLibrary/FortLewis/DiffEq/2-Higher-order/06-Forcing-resonance/KJ-3-10-08.pg
```diff
@@ -54,13 +59,18 @@ $w = 8;
 
 $F = Formula("$a e^(-t)")->reduce();
 
-$c1 = Compute("-$a/74");
-$c2 = Compute("($a/74 + 2*$c1)/6");
-$c3 = Compute("$a/74");
+## this does too much rounding for some seeds
+#$c1 = Compute("-$a/74");
+#$c2 = Compute("($a/74 + 2*$c1)/6");
+#$c3 = Compute("$a/74");
```
drdrew42, line 65:
```suggestion
```

drdrew42 (approved):
Thanks for providing this fix! I have suggested the removal of the old code -- it doesn't need to remain.

PR #1036, opened 2023-04-13
Fix wording of ellipse problem

d-torrance:
Depending on the random numbers chosen, there is a chance that the axis along the x-axis is the minor axis and not the major axis.  We fix the wording of the problem to reflect this.


drdrew42 (approved):
Excellent, thanks!

PR #1059, opened 2023-07-11
New files added

mathriddle:
Integration techniques written for APEX Calculus online text but useful for any text. Most of these require students to do the setup for the integration technique rather than just asking only for the antiderivative. Scaffording is often used.


mathriddle:
Added additional problems for integration applications involving volumes of revolution (via disks and shells) and arc length. What is perhaps different with these is providing a picture of the region being revolved and asking students to first show the setup for the definite integrals before asking for the value of the integral.

drdrew42 (approved):

PR #1069, opened 2023-07-25
Update Stewart5_7_4_2a.pg

Tanaquil18:
replaced select_multiple_checkbox with parserCheckboxList


pstaabp:
This also looks like it needs to be perltidied.

Tanaquil18:
I ran perltidy and committed again!

pstaabp:
The two comments I mentioned earlier still need to be taken care of.

Tanaquil18:
Which comments?  All I saw on this one was that it needed to be perltidied.  Is there something else I need to do?

pstaabp:
Can you look at https://github.com/openwebwork/webwork-open-problem-library/pull/1069 ?

1) Line 23 should be removed
2) The SOLUTION should be switched to PGML_SOLUTION.

Tanaquil18:
I still don't see where to find those 2 comments originally, but I see that last comment and I fixed them now!

drdrew42 (approved):

PR #1071, opened 2023-07-25
Update old table for newer one in one problem

Tanaquil18:
DataTable is newer and works better in PreTeXt

### DIFF
--- a/OpenProblemLibrary/Michigan/Chap5Sec4/Q03.pg
+++ b/OpenProblemLibrary/Michigan/Chap5Sec4/Q03.pg
@@ -42,8 +42,8 @@ loadMacros(
   "PGchoicemacros.pl",
   "PGgraphmacros.pl",
   "MathObjects.pl",
+  "niceTables.pl",
   "PGcourse.pl",
-  "niceTables.pl"
 );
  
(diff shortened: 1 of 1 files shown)


OpenProblemLibrary/Michigan/Chap5Sec4/Q03.pg
```diff
@@ -42,7 +42,8 @@ loadMacros(
   "PGchoicemacros.pl",
   "PGgraphmacros.pl",
   "MathObjects.pl",
-  "PGcourse.pl"
+  "PGcourse.pl",
+  "niceTables.pl"
```
pstaabp, line 46: `PGcourse.pl` should always be the last macro in the list.

pstaabp:
Looks like you have switched to PGML in #1069 and you should do that for this problem also.

OpenProblemLibrary/Michigan/Chap5Sec4/Q03.pg
Tanaquil18: Didn't know that, thank you!

drdrew42 (approved):

PR #1080, opened 2023-09-06
Update Unit help files for PG 2.18

duffee:
Units.pm in PG 2.18 (released summer 2023) has changed the unit `A` from Angstrom to Ampere.
This commit updates affected help files.

- `A` is now `angstrom`
- **Centigrade** is now **Celsius**
- **Degrees Kelvin** is now just **Kelvin**
- `degK` is now `K`

`amp` is still used for ampere with `A` as an alias. It can be updated when more people have upgraded their PG to version 2.18.

Merge this PR when Michigan and Fort Lewis upgrade to PG 2.18.


pstaabp:
We are no longer using macros in the OPL.   They are all in PG now including the help files.  I guess this will affect users before 2.17, but this probability be deprecated/deleted in the near future.

PR #1096, opened 2023-10-12
fix Anton9_7Q5.pg

szhorvat:
This is my first contribution here, and it was not clear if I should edit the file in Contrib, in OPL, or both.

The `if($dx==1){$dx=-1}` line must come before the `$Dx=...` line, otherwise `$dx` and `$Dx` end up being inconsistent, and WeBWorK will miscalculate the answer.


drdrew42 (approved):

drdrew42:
Congrats on your first commit! Would you mind opening another for the Library/ version of this problem? Both should be fixed.

szhorvat:
See #1100.  Why are there two copies of the problems?

drdrew42:
Problems that were accepted into the OPL retained their original versions in Contrib/

Contrib did not used to be indexed for search in WeBWorK, but now that it is, it may be undesirable to retain duplicated problems.

PR #1097, opened 2023-10-13
Clarify wording and change correct answer for TF statements

dlglin:
Tagging @jjh2b since this is taken from his textbook.

The existing problem Library/WHFreeman/Holt_linear_algebra/Chaps_1-4/4.2.41_47a.pg contains the following statements:
"If the set of vectors \(U\) is linearly independent in a subspace \( S\) then vectors can be added to \(U\) to create a basis for \(S\)"
"If the set of vectors \(U\) spans a subspace \(S\), then vectors can be removed from \(U\) to create a basis for \(S\)"

The correct answer for both of these is coded as "False", but this depends on how you interpret "can be added" and "can be removed".  If you interpret this to include the case where no vectors are added/removed, then these statements are both true.  In fact, the statements are taken directly from the textbook, and the answers in the textbook list the first statement as true (the second doesn't have an answer in the back), which contradicts the correct answer coded in the problem.

The proposed fix here is to add some more precise notation to the end of the statements.  By using `\subseteq` it removes the ambiguity as to whether the subset/superset needs to be proper.  I'd also be okay with removing the "can be added/removed" wording altogether, but that makes the statements less accessible to readers who are less mathematically sophisticated.

### DIFF
--- a/OpenProblemLibrary/WHFreeman/Holt_linear_algebra/Chaps_1-4/4.2.41_47a.pg
+++ b/OpenProblemLibrary/WHFreeman/Holt_linear_algebra/Chaps_1-4/4.2.41_47a.pg
@@ -62,24 +62,19 @@ $tf->rf_print_q(~~&pop_up_list_print_q);
 $tf->ra_pop_up_list( 
 [ No_answer => "?", "T"=>"True", "F"=>"False"] ); 
 
-#$a = random(2,3,1);
-#$aa = $a+1;
-#$b = random(6,9,1);
-#$c = random(7,10,1);
-
 # Questions and answers
 $tf -> qa ( 
 "If  \(\  S_1  \)  and \(\  S_2  \) are subspaces of \( R^n\) of the same dimension, then \(S_1=S_2\).",
 "F",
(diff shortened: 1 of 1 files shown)


gajennings:
I agree that the problem is poorly worded.  S is a subspace of what?  A finite dimensional vector space?  A topological space?...
How about this?
"Let S be a subspace of R^n, and U be a set of vectors in S."
"If the vectors in U are linearly independent, then S has a basis that contains *all* the vectors in U (and perhaps some additional vectors)". (True).
"If the vectors in U are linearly independent, then S has a basis that contains *only* vectors in U (but not necessarily all of U)." (False).
"If U is a subset of S, and U spans S, then S has a basis that contains *all* the vectors in U" (and perhaps some additional vectors).  (False)
"If U is a subset of S, and U spans S, then S has a basis that contains *only* vectors in U (but not necessarily all of U)".  (True)
then include the two questions about dimension from the original problem.

gajennings:
I added some changes to yours and converted everything to PGML, added a solution, and pushed everything to the OPL, but neglected to close the pull request.  Am doing that now.

PR #1136, opened 2024-03-10
Use "inf" instead of "infinity" for function composition intervals

d-torrance:
This way, these problems will accept input from the MathQuill editor.

### DIFF
--- a/OpenProblemLibrary/Rochester/setAlgebra17FunComposition/srw2_8_1.pg
+++ b/OpenProblemLibrary/Rochester/setAlgebra17FunComposition/srw2_8_1.pg
@@ -31,7 +31,6 @@ loadMacros(
   "PGcourse.pl"
 );
 
-TEXT(beginproblem());
 $showPartialCorrectAnswers = 1;
 
 $a = random(2,15,1);
--- a/OpenProblemLibrary/Rochester/setAlgebra17FunComposition/srw2_8_1.pg
+++ b/OpenProblemLibrary/Rochester/setAlgebra17FunComposition/srw2_8_1.pg
@@ -56,14 +56,14 @@ $BBOLD Note: $EBOLD See \{helpLink("interval")\} for help entering intervals.
 END_TEXT
 
 Context("Interval");
-$ans1 = Compute("x**2-($a-1)*x+$b");
+$ans1 = Compute("x**2-($a-1)*x+$b")->reduce();
 $ans2 = Compute("(-inf, inf)");
-$ans4 = Compute("x**2-($a+1)*x-$b");
+$ans4 = Compute("x**2-($a+1)*x-$b")->reduce();
 $ans5 = Compute("(-inf, inf)");
-$ans7 = Compute("(x**2-$a*x)*(x+$b)");
+$ans7 = Compute("(x**2-$a*x)*(x+$b)")->reduce();
 $ans8 = Compute("(-inf, inf)");
-$ans10 = Compute("(x**2-$a*x)/(x+$b)");
-$ans11 = Compute(-$b);
+$ans10 = Compute("(x**2-$a*x)/(x+$b)")->reduce();
+$ans11 = Compute("(-inf,-$b)U(-$b,inf)");
 ANS($ans1->cmp);
 ANS($ans2->cmp);
 ANS($ans4->cmp);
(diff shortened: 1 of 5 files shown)


dpvc:
I think these should all be converted to MathObjects, where both `inf` and `infinity` mean the same thing, and you will get better error messages for wrong answers.  Just my 2 cents.

d-torrance:
Sounds good!  I've adapted each of the problems to use MathObjects.

dlglin:
There are a number of other things that could (should?) be done to bring these problems up to modern standards:

- `TEXT(beginproblem());` can be removed.  It no longer does anything.
- `TEXT(EV2(<<EOT));`/`EOT` should be replaced with `BEGIN_TEXT`/`END_TEXT`.  Better yet the problem text should be converted to PGML.
- The current answer blank structure telegraphs the answers.  If it asks for a single interval then the answer is always (-inf,inf).  If it asks for the non-permissible values then it only accepts one number.  I'd suggest using the interval context for all domains and replacing the instructions with `helpLink('interval')`.

Let me know if you want me to code any of these changes.

d-torrance:
I've pushed a few new commits addressing the suggestions, except I didn't convert them to PGML since that would require a bit more work.

OpenProblemLibrary/Rochester/setAlgebra17FunComposition/srw2_8_1.pg
```diff
@@ -27,56 +27,50 @@ DOCUMENT();        # This should be the first executable line in the problem.
...
-ANS(num_cmp($ans11));
+Context("Interval");
+$ans1 = Compute("x**2-($a-1)*x+$b");
+$ans2 = Compute("(-inf, inf)");
+$ans4 = Compute("x**2-($a+1)*x-$b");
+$ans5 = Compute("(-inf, inf)");
+$ans7 = Compute("(x**2-$a*x)*(x+$b)");
+$ans8 = Compute("(-inf, inf)");
+$ans10 = Compute("(x**2-$a*x)/(x+$b)");
+$ans11 = Compute(-$b);
```
dlglin, line 66:
```suggestion
$ans1 = Compute("x**2-($a-1)*x+$b")->reduce();
$ans2 = Compute("(-inf, inf)");
$ans4 = Compute("x**2-($a+1)*x-$b")->reduce();
$ans5 = Compute("(-inf, inf)");
$ans7 = Compute("(x**2-$a*x)*(x+$b)")->reduce();
$ans8 = Compute("(-inf, inf)");
$ans10 = Compute("(x**2-$a*x)/(x+$b)")->reduce();
$ans11 = Compute("(-inf,-$b)U(-$b,inf)");
```

OpenProblemLibrary/Rochester/setAlgebra17FunComposition/sw4_7_31.pg
```diff
@@ -29,60 +29,52 @@ DOCUMENT();        # This should be the first executable line in the problem.
...
-ANS(str_cmp($ans12));
+Context("Interval");
+$ans1 = Compute("(x + $b)**2+$a");
+$ans2 = Compute("(-inf, inf)");
+$ans4 = Compute("x**2 + $a+$b");
+$ans5 = Compute("(-inf, inf)");
+$ans7 = Compute("(x**2 + $a)**2+$a");
+$ans8 = Compute("(-inf, inf)");
+$ans10 = Compute("x + 2*$b");
+$ans11 = Compute("(-inf, inf)");
```
dlglin, line 69:
```suggestion
$ans1 = Compute("(x + $b)**2+$a")->reduce();
$ans2 = Compute("(-inf, inf)");
$ans4 = Compute("x**2 + $a+$b")->reduce();
$ans5 = Compute("(-inf, inf)");
$ans7 = Compute("(x**2 + $a)**2+$a")->reduce();
$ans8 = Compute("(-inf, inf)");
$ans10 = Compute("x + 2*$b")->reduce();
$ans11 = Compute("(-inf, inf)");
```

dlglin (changes requested):
The `reduce` calls will combine some of the constants and make the correct answers look nicer.  Also there is one answer that you missed converting to an interval.

dlglin:
Also, thanks for doing the extra work on the problems.

dlglin (approved):

pstaabp (approved):
Looks good.  I would agree that this should be converted to PGML, but is still a nice improvement.

PR #1138, opened 2024-03-22
Updating problems with bugs in UBC STAT443

evsi8432:
Hi there,

I am a TA for STAT 443 at UBC, and we have some problems that had some bugs in them. I fixed the bugs, but now I need to update the repo with my changes! Let me know if you need anything from me.

Thanks,

Evan


drdrew42 (approved):
Thanks!

PR #1140, opened 2024-03-30
Production

Alex-Jordan:
This is several years of accumulated changes to PCC files in the OPL. Every change here is either in `OpenProblemLibrary/PCC`, `OpenProblemLibrary/macros/PCC`, or `Contrib/PCC`.


drdrew42:
😱 
Screenshot 2024-04-10 at 12 45 35 PM

Is this the GitHub equivalent of a trust-fall? 😆

Alex-Jordan:
Hmm... so, is there a git way to confirm that the changes here only affect PCC files?

Let me look into this on my end. I'm surprised we have that many files, but maybe it's around 2500 and then doubled for the OPL and Contrib. Maybe it will help if I split into to PRs based on that. Or chunk it out by some folders versus other folders.

drdrew42:
I just want to confirm that this is indeed limited to PCC files -- 5k is pretty massive. If it isn't too much work, splitting this into PRs at least for OPL/PCC vs Contrib/PCC would be a good start. 👍

pstaabp:
@Alex-Jordan  it would be good to get this in this summer.  Can you make some smaller PRs?

PR #1141, opened 2024-04-05
Use MathObject intervals for interval of convergence problems

d-torrance:
One of the problems was expecting the strings "-infinity" and "infinity", which doesn't work with the MathQuill interface and students were being marked wrong.

This is similar to #1136.


d-torrance:
I just added a new commit randomizing the power series in each problem a bit more variety.

drdrew42 (approved):
Thanks for these updates!

Just a note, since you're removing the now-unnecessary `TEXT(beginproblem());` -- one can also remove `$showPartialCorrectAnswers=1;` as this is the default and should only appear when disabling partial credit.

d-torrance:
> Just a note, since you're removing the now-unnecessary `TEXT(beginproblem());` -- one can also remove `$showPartialCorrectAnswers=1;` as this is the default and should only appear when disabling partial credit.

Good to know!  I thought that might be the case but wasn't sure.

PR #1144, opened 2024-04-15
Remove all .def files from Contrib

dlglin:
Set definition files shouldn't be uploaded to this repository at all.  They have not been maintained, and in most cases aren't of general use.


drdrew42 (approved):
🥳

PR #1164, opened 2024-07-17
Remove mention of "contradiction" from epsilon-delta limit proofs

d-torrance:
These are direct proofs using the definition of the limit and involve no contradiction.


drdrew42 (approved):
👍

PR #1166, opened 2024-07-28
Warren Shull WebWork Calculus scaffolds

weshuu:
I am hoping to add these 16 problems to the Open Problem Library! When this is done, I may want some guidance as to the different ways these can be searched for.

### DIFF
--- a/Contrib/Wisconsin/EauClaire/calculus/AntiderivativePowerRuleIntroduction.pg
+++ b/Contrib/Wisconsin/EauClaire/calculus/AntiderivativePowerRuleIntroduction.pg
@@ -1,0 +1,327 @@
+## DESCRIPTION
+##   ****
+## ENDDESCRIPTION
+
+## KEYWORDS('Power Rule')
+
+## DBsubject('Calculus')
+## DBchapter('Derivatives')
+## DBsection('Power Rule')
+## Date('07/20/2024')
+## Author('W. Shull')
+## Institution('UWEC')
+## TitleText1('')
+## EditionText1('')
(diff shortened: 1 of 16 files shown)


dlglin:
On first look here are some suggestions:

- You should only load the macro files that you are using in the question.  Put all of the macros into one `loadMacros` call, and make sure that `PGcourse.pl` is the last macro loaded.
- Remove any commented out code that is not relevant to the problem.
- If you are using WeBWorK 2.19, you should run "Reformat the code using perltidy" from the "Format code" tab.  It will standardize the alignment of your code.
- You can remove `TEXT(beginproblem());`.  It is no longer used.

pstaabp:
I'm taking a look at these problems.  All of the comments above still hold.  Let me point out a couple of issues that I see too:
1. There are a number of errors about loading `customCSS.pl`.  I'm guessing this is your own macro.  It should be removed. 
2. Also, in `chain_rule_problem_1_of_2_composition_review.pg`, there is a custom function called: `kalturaShull`, which might be in that macro.  If it is a robust macro, you should contribute it back.  If not, perhaps you can embed it in the problem.  There is also another one called `youtube` that doesn't exist. 
3. The problems: `Contrib/Wisconsin/EauClaire/calculus/extreme_value_theorem_problem_1_of_3.pg`, `Wisconsin/EauClaire/calculus/limits_with_formulas_problem_3_of_4.pg` and `Wisconsin/EauClaire/calculus/limits_with_formulas_problem_4_of_4.pg` won't render.  It looks like a custom function again.

PR #1172, opened 2024-08-27
Rewrite ordering problem to use DraggableProof

### LINKED ISSUE #1171: Library/LoyolaChicago/Precalc/Chap4Sec1/Q52.pg
Problem Seed: 123456

For the middle popup box, this problem only accepts "?" as the solution. The correct solution is actually "0", just as the solution text states. Something is wrong with the answer verification.

I would appreciate hints on how to fix this in the problem source code.

dlglin:
This is to fix #1171

A less intrusive fix would be to add `noindex => 1` to each of the PopUps, but this problem seemed like a great candidate for `DraggableProof`.


gajennings:
Much better!

Issue #1000, opened 2022-12-09
Issues with MathQuill and answers comparing to the string "infinity"


dlglin:
This was motivated by Library/Utah/Calculus_II/set6_Indeterminate_Forms_and_Improper_Integrals/set6_pr3.pg.

If a problem is coded to explicitly look for the string "infinity" (for example using `str_cmp`) and MathQuill is enabled, then students get some counterintuitive behaviour.  When they start typing the word infinity, they get to inf and it is replaced with the infinity symbol.  In such a circumstance if they submit their answer it is marked wrong because MQ submits the string "inf".  This is not an issue with MathObjects since "inf" is an alias for "infinity" in the Numeric context.

There are some workarounds, such as instructing students to use the text mode in MQ, typing "INFINITY" in all caps or continuing to type the word infinity despite the substitution (i.e. "∞inity" is marked right), but the right approach is to convert these questions to MathObjects and work on eliminating `str_cmp` (and other outdated answer checkers) from the OPL.

I'm going to poke around the OPL to see if I can find which problems have this issue, but it's not just a simple grep since "infinity" could be stored in a variable on a different line.  I'll post what I find here.

PR #1178, opened 2024-09-07
 Rewrite using MathObjects and PGML

lahvak:
The problem uses 'cmp_str' to check for answer 'infinity'. The `cmp_str` is deprecated, and does not work well with the MathQuill editor.

In addition to rewriting in PGML, a solution was added as well.

### DIFF
--- a/OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
+++ b/OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
@@ -46,7 +46,7 @@ If [`f`] is restricted to the above domain, then the inverse function is [`g(x)=
 END_PGML
 
 BEGIN_PGML_SOLUTION
-The function [`f(x)=(x+[$a])^2`] is one-to-one on the intervals [`(-\infty, [$ans1]]`] and [`[[$ans1], \infty)`].
+The function [`f(x)=(x+[$a])^2`] is one-to-one on the intervals [`(-\infty, -[$a]]`] and [`[-[$a], \infty)`].
 The second one contains the number 100.
 
 Solving the equation [`y=(x+[$a])^2`] for [`x`] will give us two solutions: [`x = \pm\sqrt{y} - [$a]`].
--- a/OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
+++ b/OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
@@ -30,15 +30,17 @@ $showPartialCorrectAnswers = 1;
 Context("Numeric");
 
 $a = random(2,9,1);
-$ans1 = Compute("-$a");
-$ans2 = Infinity;
 $ans3 = Formula("sqrt(x)-$a");
+
+Context("Interval");
+
+$ansInt = Compute("[-$a, infinity)");
 
 BEGIN_PGML
 The function [`f(x)=(x+[$a])^2`] is not one-to-one. Choose the largest possible
 domain containing the number 100 so that the function restricted to the domain is one-to-one.
 
-The largest possible domain is [`\left[\strut\right.`][________]{$ans1},[________]{$ans2}[`\left.\strut\right)`].
+The largest possible domain is [_________________]{$ansInt} (use interval notation).
 
 If [`f`] is restricted to the above domain, then the inverse function is [`g(x)=`][_______________]{$ans3}.
 END_PGML
--- a/OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
+++ b/OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
@@ -40,7 +40,7 @@ domain containing the number 100 so that the function restricted to the domain i
 
 The largest possible domain is [`\left[\strut\right.`][________]{$ans1},[________]{$ans2}[`\left.\strut\right)`].
 
-The inverse function is [`g(x)=`][_______________]{$ans3}.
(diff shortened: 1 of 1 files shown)


lahvak:
This should fix another instance of #1000

OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
```diff
@@ -20,35 +20,37 @@ DOCUMENT();        # This should be the first executable line in the problem.
...
 $a = random(2,9,1);
+$ans1 = Compute("-$a");
+$ans2 = Infinity;
+$ans3 = Formula("sqrt(x)-$a");
+
+BEGIN_PGML
+The function [`f(x)=(x+[$a])^2`] is not one-to-one. Choose a largest possible
+domain containing the number 100 so that the function restricted to the domain is one-to-one.
+
+The largest possible domain is [`\left[\strut\right.`][________]{$ans1},[________]{$ans2}[`\left.\strut\right)`].
```
dlglin, line 41: This would be cleaner using the `Interval` context.

OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
```diff
@@ -20,35 +20,37 @@ DOCUMENT();        # This should be the first executable line in the problem.
...
 
+Context("Numeric");
+
 $a = random(2,9,1);
+$ans1 = Compute("-$a");
+$ans2 = Infinity;
+$ans3 = Formula("sqrt(x)-$a");
+
+BEGIN_PGML
+The function [`f(x)=(x+[$a])^2`] is not one-to-one. Choose a largest possible
```
dlglin, line 38: There is a unique largest possible domain containing 100, so how about changing this to "Choose the largest possible domain..."

OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
```diff
@@ -20,35 +20,37 @@ DOCUMENT();        # This should be the first executable line in the problem.
...
+domain containing the number 100 so that the function restricted to the domain is one-to-one.
+
+The largest possible domain is [`\left[\strut\right.`][________]{$ans1},[________]{$ans2}[`\left.\strut\right)`].
+
+The inverse function is [`g(x)=`][_______________]{$ans3}.
```
dlglin, line 43: This should specify that it is asking for the inverse function on the domain given above.  How about "If x is restricted to the above domain, then the inverse function is "?

OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
lahvak: Done.

OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
lahvak: I was thinking about it, but did not want to make too large changes to the problem.

OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
lahvak: Sure, that is definitely better.

OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
dlglin: IMO there is no such thing as too large of changes.  If there are places where the problem can be improved while still effectively asking the same question then I say go for it.

OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
```diff
@@ -20,35 +20,39 @@ DOCUMENT();        # This should be the first executable line in the problem.
...
-EOT
-
-$ans1 = -$a;
-$ans2 = "infinity";
-$ans3 = "sqrt(x)-$a";
-ANS(num_cmp($ans1));
-ANS(str_cmp($ans2));
-ANS(fun_cmp($ans3));
+Solving the equation [`y=(x+[$a])^2`] for [`x`] will give us two solutions: [`x = \pm\sqrt{y} - [$a]`].
+According to our domain choice, [`x \ge [$ans1]`], which means we have to choose the solution with [`+`].
```
gajennings, line 53: line 53: replace $ans1 with $a

OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
```diff
@@ -20,35 +20,39 @@ DOCUMENT();        # This should be the first executable line in the problem.
...
+
+Context("Interval");
+
+$ansInt = Compute("[-$a, infinity)");
+
+BEGIN_PGML
+The function [`f(x)=(x+[$a])^2`] is not one-to-one. Choose the largest possible
+domain containing the number 100 so that the function restricted to the domain is one-to-one.
+
+The largest possible domain is [_________________]{$ansInt} (use interval notation).
```
gajennings, line 43: line 43: replace "interval notation" with [@ helpLink("interval","interval notation") @]*
(including the star at the end!)  This will create a link that points to instructions for entering intervals into webwork.

OpenProblemLibrary/ma117DB/set5/srw2_9_63.pg
gajennings, line 53: Also line 53: I'd suggest this wording:  
  The correct sign for the square root is + because [`x \geq [$a]`] in the domain [`[-[$a],\infty)`].

PR #1179, opened 2024-09-07
Prevent a point on a graph being placed on the edge of window with invisible label

lahvak:
The problem generates two random but distinct numbers between 2 and 5.
In order to assure that they are distinct, it increases one of them by
1 if they are equal. This may result in one of the numbers being 6,
which will cause a point on the graph placed on the edge of the graphing
window, with invisible label, which makes it impossible to answer the
problem.

The fix selects the second number repeatedly until the numbers are
distinct.

Also, the problem was converted to PGML.

### DIFF
--- a/OpenProblemLibrary/Rochester/setAlgebra28ExpFunctions/ur_le_1_5.pg
+++ b/OpenProblemLibrary/Rochester/setAlgebra28ExpFunctions/ur_le_1_5.pg
@@ -30,8 +30,7 @@ TEXT(beginproblem());
 $showPartialCorrectAnswers = 1;
 
 $a = random(2,5,1);
-$y = random(2,5,1);
-if ($y == $a) {$y = $y+1;}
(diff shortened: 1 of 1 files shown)


dlglin:
A more efficient way to generate two distinct values would be to use the new `random_subset` function:
```
($a,$y) = random_subset(2,[2..5]);
```

lahvak:
> A more efficient way to generate two distinct values would be to use the new `random_subset` function:

I was not sure whether this is more efficient with only two elements in the subset, but it is definitely more readable.

Is there a way for a problem to specify the minimum pg version required?

dlglin:
> Is there a way for a problem to specify the minimum pg version required?

Not yet.  We have talked about it, but it hasn't been implemented.  I've opened #1187 to discuss this.

drdrew42 (approved):

PR #1183, opened 2024-09-13
Fix the percentages in the problem text

lahvak:
The numbers in the text of the problem are referred to as "percents", but they are actually probabilities, so the text says things like _0.07 percent_ instead of _7 percent_.

This replaces the probabilities by percents.  It also  removes unnecessary macro files and converts the problem to PGML.


drdrew42 (approved):
👍

PR #1190, opened 2024-09-19
Fix instance of hardcoded value when variable should be used

### LINKED ISSUE #1186: calculate answer, not hard-code
This line should end with not 42, but whatever the calculated value of the answer is.

https://github.com/openwebwork/webwork-open-problem-library/blob/90d5a3c221fba92e06eac86a3e4e8c917d1403c5/Contrib/DMOI/1-Counting/1_1-Additive_and_Multiplicative_Principles/1_1_2.pg#L91

Hat tip to one of my students for noticing this in DMOI 3e.

dlglin:
This should fix #1186.


drdrew42 (approved):
😂

PR #1191, opened 2024-09-20
Fix events for probability 0 or 1

lahvak:
With the directions of the inequalities, the event in the last part of the problem will always have probability 1, even if `$Cans` is 0, 
so there is a 50% chance that the part will be graded incorrectly.

This fixes that issue, and adds more versions of the event in the last part.


drdrew42 (approved):
Good catch!

PR #1218, opened 2024-12-18
Fix typos in taxonomy

dqnykamp:
This PR fixes a few typos in the taxonomy files:
- fix misspelling of "Inhomogeneous"
- correct links that were incorrectly attributed to "Cross product"
- normalize character used for dash (hyphen might not be correct, but it is consistent with the rest of the file)

### DIFF
--- a/OpenProblemLibrary/AlfredUniv/PDE/separable/nonhomogeneous3.pg
+++ b/OpenProblemLibrary/AlfredUniv/PDE/separable/nonhomogeneous3.pg
@@ -5,7 +5,7 @@
 
 ## DBsubject(Differential equations)
 ## DBchapter(Partial differential equations)
-## DBsection(Inhomogenous equations)
+## DBsection(Inhomogeneous equations)
 ## Date(9/1/2011)
 ## Institution(Alfred University)
 ## Author(Darwyn Cook)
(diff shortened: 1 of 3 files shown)


drdrew42:
Thanks for your attention to these typos.

Regarding the "Cross Product" links, good catch -- these are indeed misdirected.

Regarding the emdash/hyphen, this is also a linked category so the update shouldn't have any ramifications.

However, for the mis-spelling of 'inhomogeneous', things get more complicated -- updating the Taxonomy file is just the beginning. 

The taxonomy that appears in WeBWorK is actually constructed from the tags that appear in each problem file, and this `Taxonomy2` file merely acts as the source of truth for what is considered an 'official' subject/chapter/section. 

So updating the Taxonomy2 file will make the 'official' version of the `Inhomogeneous equations` DBsection use the correct spelling, **but** any problem that used the old tag (looks like that's only one problem, thankfully: `OpenProblemLibrary/AlfredUniv/PDE/separable/nonhomogeneous3.pg`) will no longer appear in the library browser because it still uses the mis-spelled, now-unofficial, tag.

Would you mind updating your PR to also include the updated DBsection tag in `OpenProblemLibrary/AlfredUniv/PDE/separable/nonhomogeneous3.pg`?

dqnykamp:
I updated that problem to use the new official tag.

There are 6 additional misspellings of "homogenous" in this repository (two in the actual OpenProblemLibrary and four in Contrib). I didn't change those, as I imagine that isn't my role.

PR #1231, opened 2025-01-28
Bug fix: +C not included

debbieyuster:
Inserted +C after the answer blank for this indefinite integration problem.


gajennings:
Thanks, I updated the code to use more modern techniques: MathObjects and PGML, and incorporated your code from your pull request.

PR #1279, opened 2025-06-21
AnswerFormatHelp removal from the OPL

pstaabp:
This removes AnswerFormatHelp from the OPL.  This includes both the macro from the `loadMacros` and then changes to the function call to `helpLink`.


pstaabp:
I'm going to close this and put in smaller ones.

PR #1280, opened 2025-06-24
Update QR related OPL problems

pstaabp:
Converted to PGML, images to tikz and add solutions for problems used in a QR course.


gajennings:
Please ignore my previous comment.  I can get the list of files from a diff.

Issue #1289, opened 2025-08-29
Library/Michigan/precalc/4e/Chap1Sec2/Q15.pg


katievicjo:
These are the warnings that I'm receiving:

> Warning messages

    Non fatal warnings. These are only displayed for users with permission to view problem debugging info.
    Use of uninitialized value in concatenation (.) or string at line 14 of (eval 1616)
    from within main::EV3 called at line 87 of Library/Michigan/precalc/4e/Chap1Sec2/Q15.pg
    Use of uninitialized value in concatenation (.) or string at line 14 of (eval 1616)
    from within main::EV3 called at line 87 of Library/Michigan/precalc/4e/Chap1Sec2/Q15.pg

Any advice?

Problem Seed: 408

dlglin:
This is due to old syntax being used in the solution.  I have submitted an update to the problem at #1290, which should be incorporated into the next release.  In the meantime you can safely use the problem as-is with students, as these warnings are only shown to instructors, and in this case the error only affects the layout and not the content.

katievicjo:
Thank you!

PR #1290, opened 2025-08-29
Convert problem to PGML

dlglin:
This is to address #1289


gajennings:
Suggestion: drop line 50: 
    $gdisp    = FEQ("$b + $m x");
and on line 61 replace $gdisp with $g.  It's a Formula object so it should display correctly.

PR #1295, opened 2025-09-16
propose fix some typos

jeis4wpi:
propose these changes


dlglin (approved):
One thing to be aware of: if you change the taxonomy, then any problems that are classified with the modified DBsubject, DBchapter or DBsection will need to be edited to match.

This is a non-issue in this case since there are no problems in the OPL using the modified tags.

PR #1312, opened 2025-10-23
bug fixes for solution to Michigan/Chap14Sec7/Q21.pg

davidlippel:
The existing code for the solution throws a non-fatal warning for privileged users (uninitialized variable `@fyyExpl`) and present an incorrect/incomplete solution to student users.

Details of my fixes:
- added some missing parentheses around ternary conditionals
- the select logic for the `fvvExpl` arrays is incorrect: we want to switch on f_vv*f_v > 0 (not f_vv > 0)
- typo on line 126: we want to initialize `@fyyExpl`, not `@fyExpl`

### DIFF
--- a/OpenProblemLibrary/Michigan/Chap14Sec7/Q21.pg
+++ b/OpenProblemLibrary/Michigan/Chap14Sec7/Q21.pg
@@ -5,7 +5,6 @@
 
 ## Tagged by glr 01/11/10
 
-
 ## DBsubject(Calculus - multivariable)
 ## DBchapter(Differentiation of multivariable functions)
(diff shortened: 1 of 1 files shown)


dlglin:
Thanks for submitting this.  Would you mind making a few additional updates to the problem:

- Remove line 134.  The `beginproblem()` method is no longer used.
- Change line 184 to `BEGIN_SOLUTION` and remove line 185.
- If you are using WW 2.20, please run "Reformat the code using perltidy" in the "Format Code" tab of the problem editor.

Ideally the problem should be converted to PGML.  It would be great if you're willing to do that, but it's more work.

dlglin (approved):

PR #1355, opened 2025-12-12
alt-text format and use shorter alt-text in Michigan/Chap2Sec3/Q41

Tanaquil18:


gajennings:
Looks good.  Thanks again!

PR #1357, opened 2025-12-15
Fix formatting of integral answer in sc5_5_24.pg add + C

ggarza:
Several students were unsure whether the answer should include a “+ C.” In its current form, the problem marks responses that include “+ C” as incorrect, which creates unnecessary ambiguity. This change resolves that issue by explicitly including “+ C” in the problem statement, making the expectation clear and consistent with how similar problems are presented elsewhere in WeBWorK.


gajennings:
It's fixed now.

PR #1359, opened 2025-12-16
Alt Text descriptions

nelong:
Updated Alt Text and Long_descriptions for exercises used in Active Calculus - Multivariable 2nd Edition.


OpenProblemLibrary/Michigan/Chap12Sec4/Q15.pg
gajennings: Please update your copy of the Open Problem Library before attempting to modify this file.  This file was updated 3 months ago in the OPL and some of the changes here undo improvements that were made in the updated version.  For example the updated version uses PGML but the version in this pull request does not.

gajennings (changes requested):

PR #1363, opened 2025-12-16
Add alt text in Rochester/setDerivatives1/nsc2s10p1_mo.pg

Tanaquil18:


Tanaquil18:
Wow, you merged these faster than I do within my own projects! :-) Thank you!

Chrissy

PR #1385, opened 2026-01-06
One more Leibniz correction please

jeis4wpi:
One more Leibniz correction please


OpenProblemLibrary/OPL_global_statistics.sql
gajennings: I don't think you meant to include this huge file in your pull request.  Maybe it got pulled in by mistake if you  ran "git add *" because that will pull in everything that has changed.  Try running "git add *.pg" to avoid pulling in things that aren't problem files.  OPL_global_statistics.sql just stores statistical information about the usage of webwork problems.  Follow this link for a fuller explanation
https://wiki.openwebwork.org/wiki/OPL_Problem_Statistics

jeis4wpi:
I am so glad you pointed out that the file move might break some installs.  I didn’t think of that.  That is the last thing I would want to do.

Thanks


  *
John

________________________________

PR #1408, opened 2026-02-06
Improve wording for ambiguous question

oscarlevin:
One of the parts of this problem from DMOI had ambiguous wording.  This fixes that.

### DIFF
--- a/Contrib/DMOI/1-Counting/1_1-Additive_and_Multiplicative_Principles/1_1_2.pg
+++ b/Contrib/DMOI/1-Counting/1_1-Additive_and_Multiplicative_Principles/1_1_2.pg
@@ -72,7 +72,7 @@ a.  You realize that the interview is for clown college, so you should probably 
     [_____]{$ans2}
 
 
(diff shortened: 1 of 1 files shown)


Contrib/DMOI/1-Counting/1_1-Additive_and_Multiplicative_Principles/1_1_2.pg
```diff
@@ -72,7 +72,7 @@ a.  You realize that the interview is for clown college, so you should probably
     [_____]{$ans2}
 
 
-a.  For the rest of your outfit, you have [$c] shirts, [$d] skirts, [$f] pants, and [$g] dresses. You want to select either a shirt to wear with a skirt or pants, or just a dress. How many outfits do you have to choose from?
+a.  For the rest of your outfit, you have [$c] shirts, [$d] skirts, [$f] pants, and [$g] dresses. You want to select either a shirt to wear with a skirt or pants, or just a dress. How many outfits do you have to choose from (not including the choice of ties)?
```
gajennings: This part of the problem refers to "outfits" where it really only means "rest of the outfit".  Would it be even clearer to make that explicit by asking "How many possibilities are there for the rest of your outfit (not including the tie)?"?

PR #1431, opened 2026-03-31
Guided/scaffolded Problems

mehdimoodi:
These problems were developed for introductory college-level probability/statistics courses and are designed to support step-by-step learning through scaffolded sections, conceptual checkpoints, and final-answer validation.


pstaabp:
It would good if these were converted to PGML, which is currently the standard.  If you have 2.20 or have switched to 2.21, then it is built into PG editor, there is a PGML converter.

PR #1438, opened 2026-05-29
Make sure foil randomization cannot be correct.

somiaj:
It is possible that one of the foils produces the correct answer. This will randomize values in a loop until the foil is guaranteed to not give the correct answer.

A second issue was brought to my attention. One of the draggable proof problems had the same statement listed twice. This was because the wrong variable was being used for the extra statement. I believe the problem should have used the incorrect statement for the extra incorrect statements.


somiaj:
@oscarlevin I think these are your problems, so hoping you can double check I fixed them properly.

PR #1445, opened 2026-07-29
Update two UMich double integral problems

dlglin:
Chap16Sec2/Q23.pg: Convert to PGML; change sketch of region of integration from multiple choice to using the graph tool to sketch the graph.

Chap16Sec7/Q25.pg: Convert to PGML; change `\int` to `\iint` where appropriate.

Tagging @glarose in case he has thoughts about either of these changes.  If I don't hear back, I'll just merge the changes.


pstaabp:
These look good to me.  Only suggestion would be to put the equations in the solution of the first problem in a single 
` [```   ```]` block and then align the equations with the `aligned` latex environment.

PR #1440, opened 2026-06-03
Problems from Louisiana Tech

jpaulterry:
Added a folder for a discrete math course along with several problems organized by sections.

PR #1447, opened 2026-08-04
updates to LA tech problems

pstaabp:
This is some minor fixes to those problems in #1440.


drgrice1:
It seems that this also fixes the CRLF line endings that is done in #1452.  So this pull request could also be merged as an alternate to that one.  Assuming that the other changes are acceptable.

PR #1452, opened 2026-08-13
Renormalize line endings in recrel1-4.pg to match .gitattributes

drgrice1:
These files were added via GitHub web upload, which bypassed the clean filter and left CRLF line endings in the stored blobs despite the *.pg text rule in .gitattributes. That mismatch made git report them as perpetually modified.

Do not use direct GitHub file uploads to the main branch.  @jpaulterry did this.  DO NOT  EVER DO THIS AGAIN.  It leaves the repository in a dysfunctional state.

This should be reviewed and merged as quick as possible.


jpaulterry (approved):

jpaulterry:
My apologies. I was unaware that the web upload feature was dangerous.

drgrice1:
Yeah, it circumvents the usual protections on the repository that a pull request goes through.  I changed the settings on this repository and I think that direct file upload should not be allowed now.

It was quite odd.  I pulled the changes, and these four files were reported as changed by git.  I tried `git restore` and `git checkout` on the files, and they remained changed.  I couldn't change branches or anything until I edited the files, converted the CRLF endings to LF endings  and committed the changes (basically this pull request).  I can't switch back to main again until this is merged.  Others are going to have the same problem (perhaps MS Windows users won't?).  We are probably going to have to educate many on how to fix their repository just so that they can pull main again after this is merged.

dlglin (approved):

